PR #3, opened 2016-10-25
Update to test_track_js_client 1.3.6

chrislopresto:


jmileham:
LGTM! Worth cutting a version of the rails client or should we wait and roll up some more change?

bower.json
```diff
@@ -12,6 +12,6 @@
     "tests"
   ],
   "dependencies": {
-      "test_track_js_client": "https://github.com/Betterment/test_track_js_client.git#v1.3.3"
+      "test_track_js_client": "https://github.com/Betterment/test_track_js_client.git#v1.3.6"
```
samandmoore: We can swap this to pull from bower, right? Or, we still have to publish the js client to bower. I will add that as an issue on the js repo.

aburgel:
bump on this. we also need to run `rake vendor_deps` and commit the minified js client

aburgel:
#13 upgraded the js client to 1.3.7. So this is obsolete.

bower.json
chrislopresto: I'd say yes. Let's publish to bower as part of our release process.

PR #5, opened 2016-10-27
Call out to TestTrack README about seed apps

jmileham:
/domain @samandmoore 

Companion to https://github.com/Betterment/test_track/pull/4

The link won't autoscroll until that ships.


samandmoore:
<<domainlgtm

oh hi, nanda!

PR #7, opened 2016-10-31
Update generators with experiment/gate/drop flags

while1malloc0:


README.md
```diff
@@ -107,6 +107,70 @@ end
 
 _Note: `drop_split` (a.k.a. `finish_split`) does not physically delete split data from mixpanel or Test Track's database._
 
+### Generating Split Migrations
+
+Split configuration changes can be generated using Rails generators that are included with the TestTrack Rails client.
+
+```
+rails generate test_track:migration add_name_of_split
```
jmileham: `add_[name_of_feature]` might be a good way to talk about this instead of `name_of_split` if we're going to be suffixing it with either `_experiment` or `_enabled` if we're given a type?

Also ActiveRecord migrations can infer type from the name of the migration - maybe you can get the same effect as `-t drop` by calling with `drop_[name_of_feature]`, or the same effect as `-t experiment` by calling with `add_[name_of_feature]_experiment`.

README.md
```diff
@@ -107,6 +107,70 @@ end
...
+class AddNameOfSplit < ActiveRecord::Migration
+  def change
+    TestTrack.update_config do |c|
+      c.split :name_of_split, control: 100, treatment: 0
+    end
+  end
+end
+```
+
+Different types of splits can be generated using the `--type/-t` flag with the options of `experiment`, `gate` (a feature flag), or `drop` (ending a split)
```
jmileham: Maybe we should decide that the suffix for gates should be `_gate` instead of `_enabled` (i know, I'm on the record saying that _enabled is a good convention, but wanted to talk a bit more about it)? I do like that intuitively `_enabled` has an obviously boolean set of values, but the words gate and enabled aren't super obviously related. /cc @swiftsam

README.md
while1malloc0: Yeah, that was the original approach and for some reason I thought that the flag approach might be better. I can definitely see the utility in having this act like regular rails migrations, so I'll update this feature accordingly.

README.md
while1malloc0: Based on the comment above, if we use the name of the split to infer its meaning, then perhaps it makes more sense here to have a name that ends in `_enabled` or `Enabled` create a feature flag split, since the only difference there would be in the names of the variants.

jmileham: yeah, then we can dodge the question of what we call feature flags/gates entirely, and just go after _enabled as a boolean thing that defaults 100% false and happens to be kind of ideal for feature gating.

swiftsam: yeah, I think `enabled` is more meaningful than `gate` in general, so I think I agree with where you guys ended up to use `_enabled` and `_experiement` as our suffixes

while1malloc0:
@jmileham See what you think about this approach

lib/generators/test_track/migration_generator.rb
```diff
@@ -1,38 +1,77 @@
...
+      def split_class_name
+        split_name.camelize
+      end
+
+      def full_file_path
+        "db/migrate/#{formatted_time_stamp}_#{split_name}.rb"
+      end
+
+      def split_command_line
+        "#{split_command} :#{split_name.gsub('finish_', '')}#{split_variants}"
```
jmileham: `split_name` doesn't seem like the right name for this method if it potentially has `finish_` on the front, right? Maybe more methods that do more different slicing and dicing?

lib/generators/test_track/migration_generator.rb
```diff
@@ -1,38 +1,77 @@
...
+            ', control: 100, treatment: 50'
+          else
+            ', control: 50, treatment: 50'
+        end
       end
 
-      def finish_split?
-        file_name.start_with?('retire', 'finish')
+      def split_type
+        if split_name.start_with? 'finish', 'retire', 'drop'
```
jmileham: We should probably just allow `drop` unless this is inspired by ActiveRecord migrations

lib/generators/test_track/migration_generator.rb
jmileham: We should probably also treat `create_` and `update_` as noisewords or something, but those noisewords might want to go into the filename.

lib/generators/test_track/migration_generator.rb
while1malloc0: I was just trying to allow for a few different ways of phrasing that, but if we want limit that to one canonical way to drop splits that sounds good too. I'll remove the additional ones.

while1malloc0:
I think this is looking a bit better now. I'm denoising the split name a bit more and have some more fine-grained methods for permuting the file name into different configurations.

while1malloc0:
Closing this in order to do some better branch management on my fork.

Issue #9, opened 2016-11-03
Re-work SafeWrapper default_handler to log.error instead of Airbrake


samandmoore:
The errors that we catch inside of [`SafeWrapper`](https://github.com/Betterment/test_track_rails_client/blob/master/app/models/test_track/analytics/safe_wrapper.rb) are from analytics client libraries. As such, it doesn't make very much sense to send them over to Airbrake as they are likely to be temporary runtime failures (e.g. http connection issues) and not fixable software bugs.

Also, given that we have the ability to specify our own custom `error_handler` (in the analytics `SafeWrapper`) via a lambda, there is no need to keep the Airbrake code in the default configuration. If someone really wants to log errors coming out of their analytics library integration, then they can configure the `error_handler` on their custom analytics client.

_edited: to add clarification_

ceslami:
👋 

Very interested in the analytics library aspect of this, as timeouts in the Mixpanel client have been frequenting our error tracker. 

Do you think it should be coupled to removing Airbrake?

jmileham:
So you're not calling for removal of Airbrake support in general, are you? Because many uses of Airbrake are to alert on software quality issues in the consuming app code - failing to use `#vary` or `#ab` correctly in ways that we want to gloss over for the end user's benefit (i.e. don't break the UX just because the app code is inconsistent with the split configuration). I spose we could lose the airbrake coupling for those use cases by simply raising in a new thread and letting whatever error notifier you have Just Work. Seems a little weird, but maybe that'd be a way to eliminate that explicit coupling.

ceslami:
To clarify: I don't have a particular opinion whether we should remove Airbrake, but I definitely think we should prevent our analytics clients from raising.

jmileham:
We do prevent them from raising, that's actually what SafeWrapper does. We just also Airbrake explicitly as a side effect, which is what @samandmoore is proposing pulling.

ceslami:
Got it -- my bad.

ceslami:
Re: analytics clients raising: at one point I thought we decided to log when there are analytics client errors. We use Airbrake in other places, and maybe that usage is subject to conversation, but we might already have our answer to how we want to handle analytics library errors.

ceslami:
Sorry for hijacking the thread -- I think we're on the same page

samandmoore:
Sorry if that last bit of the issue was unclear.

[...]

PR #12, opened 2016-11-03
Prevent analytics clients from notifying via Airbrake

ceslami:
Addressing: https://github.com/Betterment/test_track_rails_client/issues/9

Failures in analytics clients tend to be temporary runtime/network
failures and not an indication that there is broken code to
fix. As a remedy, we *only* log in the default error_handler
proc in SafeWrapper, which wraps all analytics clients.

/domain @samandmoore @jmileham @aburgel @rynonl


spec/dummy/tmp/test_track_schema.yml
```diff
@@ -0,0 +1,13 @@
...
+- b
+- c
+splits:
+  balance_unit:
+    dollar: 50
+    pound: 25
+    doge: 25
+  blue_button:
+    'false': 50
+    'true': 50
```
ceslami: Hmm -- maybe we don't want to commit this?

rynonl:
<< domainlgtm

spec/dummy/tmp/test_track_schema.yml
rynonl: oh hrm, browsed right over this.  yea maybe gitignore everything but the .keep file.

ceslami:
There's also the `spec/dummy/tmp/cache` directory, which if not present, breaks the tests. After I run `bundle exec rake`, it empties that directory, including the `.keep` file, so it gets left out. Without that directory, a clean pull of this branch will fail until the developer adds it back.

Any ideas? Couldn't find the code/config that explicitly empties that directory out after `rake` runs.

rynonl:
there's a `mkdir spec/dummy/cache` in the `before` block of the travis config. Seems reasonable to add(or move) that to a `before :spec` hook in the rake file

.gitignore
```diff
@@ -5,7 +5,7 @@ pkg/
 spec/dummy/db/*.sqlite3
 spec/dummy/db/*.sqlite3-journal
 spec/dummy/log/*.log
-spec/dummy/tmp/
```
jmileham: you want the `spec/dummy/tmp/` (in fact, it'd be appropriate to prefix all these spec lines with a `/` - you then want to explicitly allow .keep with a `.keep` line.

jmileham:
/platform @jmileham
<< platform

.gitignore
jmileham: See https://github.com/Betterment/test_track_chrome_extension/pull/5/files

jmileham:
tafn - just want to get the `.gitignore`ing right.

.gitignore
jmileham: This still needs to be in here

jmileham:
platform lgtm!

ceslami:
Ha! Beat me to the bump!

ceslami:
Thanks for your help with the cache clearing issue @jmileham 

Nice outcome of this: contributors will have a much better first-time experience!

PR #14, opened 2016-11-13
Use bower instead of github for js client

### LINKED ISSUE #4: Use the js client from bower not GitHub
We're currently pulling the JS client in via bower but using a git reference instead of the reference to the bower package in the bower registry. 

We should swap our config to use the proper bower dependency once we have the package in the bower registry.

aburgel:
/domain @rynonl @samandmoore 

Fixes #4

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -12,6 +12,6 @@
     "tests"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


samandmoore:
<<domainlgtm! 👍

PR #15, opened 2016-11-16
Update generators

while1malloc0:
Same as https://github.com/Betterment/test_track_rails_client/pull/7

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ will generate a timestamped migration file with the content
 class AddNameOfSplit < ActiveRecord::Migration
   def change
     TestTrack.update_config do |c|
-      c.split :name_of_split, control: 100, treatment: 0
+      c.split :name_of_split, control: 50, treatment: 50
     end
   end
 end
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ will generate a timestamped migration file with the content
 class AddNameOfSplit < ActiveRecord::Migration
   def change
     TestTrack.update_config do |c|
-      c.split :name_of_split, control: 50, treatment: 50
+      c.split :name_of_split, control: 100, treatment: 0
     end
   end
 end
@@ -132,7 +132,7 @@ The generator infers the type of split from the migration name.
 Adding `Drop` to the migration name will create a migration to drop a split.
 
 ```
-rails generate test_track:migration DropNameOfSplit
+rails generate test_track:migration drop_name_of_split
 ```
 
 ```ruby
@@ -148,7 +148,7 @@ end
 Adding `Enabled` or `FeatureFlag` to the end of the migration name will create sensible defaults for a feature flag split.
 
 ```
-rails generate test_track:migration AddNameOfSplitEnabled
+rails generate test_track:migration add_name_of_split_enabled
 ```
 
 ```ruby
@@ -164,14 +164,14 @@ end
 Adding `Experiment` to the end of the migration name will create sensible defaults for an experiment.
 
 ```
-rails generate test_track:migration AddNameOfSplitExperiment
+rails generate test_track:migration add_name_of_split_experiment
 ```
 
 ```ruby
 class AddNameOfSplitExperiment < ActiveRecord::Migration
   def change
     TestTrack.update_config do |c|
-      c.split :name_of_split_experiment, control: 100, treatment: 0
+      c.split :name_of_split_experiment, control: 50, treatment: 50
     end
   end
 end
--- a/README.md
+++ b/README.md
@@ -127,10 +127,10 @@ class AddNameOfSplit < ActiveRecord::Migration
 end
 ```
 
-Different types of splits can be generated using the `--type/-t` flag with the options of `experiment`, `gate` (a feature flag), or `drop` (ending a split)
+The generator infers the type of split from the split name. For instance, adding `Drop` to the beginning of a split name will generate a migration to drop a split.
 
 ```
-rails generate test_track:migration name_of_split -t drop
+rails generate test_track:migration DropNameOfSplit
 ```
 
 ```ruby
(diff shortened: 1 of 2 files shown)


lib/generators/test_track/migration_generator.rb
```diff
@@ -1,38 +1,82 @@
...
-      def split_command
-        @split_command ||= finish_split? ? 'c.finish_split' : 'c.split'
+      def split_class_name
+        split_file_name.camelize
       end
 
-      def finish_split?
-        file_name.start_with?('retire', 'finish')
+      def split_name
+        noise_words = /finish_|create_|update_|drop_/
```
rynonl: `add_`

lib/generators/test_track/migration_generator.rb
```diff
@@ -1,38 +1,82 @@
...
+            ', true: 0, false: 100'
+          when :experiment
+            ', control: 100, treatment: 50'
+          else
+            ', control: 50, treatment: 50'
+        end
+      end
+
+      def split_type
+        if split_file_name.start_with? 'drop'
```
rynonl: || 'finish'?

while1malloc0:
re: the readme, that still needs to be updated. Was going to do that once the code changes were solidified.

lib/generators/test_track/migration_generator.rb
rynonl: That's fine too, but should probably remove `finish_` from the noise words.

while1malloc0:
bump @rynonl @jmileham

lib/generators/test_track/migration_generator.rb
jmileham: Bump on this - this file should totally remove the word finish I think - finish is a deprecated method name that only continues to exist so that migrations in the wild don't break.

lib/generators/test_track/migration_generator.rb
```diff
@@ -1,38 +1,82 @@
...
-      def split_command
-        @split_command ||= finish_split? ? 'c.finish_split' : 'c.split'
+      def split_class_name
+        split_file_name.camelize
       end
 
-      def finish_split?
-        file_name.start_with?('retire', 'finish')
+      def split_name
+        noise_words = /^create_|^update_|^drop_/
```
jmileham: The documentation above uses `... add_name_of_split` but this implementation won't result in the example migration.

jmileham:
<< platform domain tafn

lib/generators/test_track/migration_generator.rb
while1malloc0: Got it, I thought that we only wanted to deprecate its use as a noise word. I changed all references to `drop`.

lib/generators/test_track/migration_generator.rb
while1malloc0: Derp. I updated the readme with examples and added `add` back in as a noise word, since I think that the probability of starting a proper split name with `Add` is probably low.

while1malloc0:
bump

rynonl:
<< domainlgtm

not gonna block on it, but want to point out that this is broken given our current naming practices

AddMyCoolSplitQ42016Enabled -> :my_cool_split_q42016_enabled instead of :my_cool_split_q4_2016_enabled

while1malloc0:
re: breaking our naming conventions, I'm not sure that we want to try and
fix that since other people using TT might have different naming
conventions wrt how the they track time etc.

jmileham:
I always call generators snake case anyway, so @rynonl's point shouldn't matter in that case, right?

README.md
```diff
@@ -107,6 +107,76 @@ end
...
+      c.drop_split :name_of_split
+    end
+  end
+end
+```
+
+Adding `Enabled` or `FeatureFlag` to the end of the migration name will create sensible defaults for a feature flag split.
+
+```
+rails generate test_track:migration AddNameOfSplitEnabled
```
jmileham: Maybe to avoid the issue @rynonl mentioned, you should just encourage snake case calls in these two examples

README.md
```diff
@@ -107,6 +107,76 @@ end
...
+
+```
+rails generate test_track:migration AddNameOfSplitExperiment
+```
+
+```ruby
+class AddNameOfSplitExperiment < ActiveRecord::Migration
+  def change
+    TestTrack.update_config do |c|
+      c.split :name_of_split_experiment, control: 100, treatment: 0
```
jmileham: Dont' these default to 50/50?

jmileham:
just those two nits, and platformlgtm!

jmileham:
domain lgtm

README.md
while1malloc0: For some reason I was thinking experiment = 100/0 and "other" splits = 50/50, but I double checked and 50/50 is correct. I updated that here and in the readme.

README.md
while1malloc0: That works. I always use camelcase for migration generators as do the rails docs. But, they don't have the this problem :-)

lib/generators/test_track/migration_generator.rb
```diff
@@ -1,38 +1,82 @@
...
+      def split_variants
+        case split_type
+          when :drop
+            ''
+          when :gate
+            ', true: 0, false: 100'
+          when :experiment
+            ', control: 50, treatment: 50'
+          else
+            ', control: 100, treatment: 0'
```
jmileham: what is the value in this fork? I don't think we'd want a 100% control situation in general - it implies you'll be changing the weightings later which would harm analysis. I think this should just line up with `:experiment` behavior.

README.md
jmileham: They actually have the same problem around any ambiguous underscore conversion 'cause they make table snake case table names.

lib/generators/test_track/migration_generator.rb
while1malloc0: sg. Implemented that in https://github.com/Betterment/test_track_rails_client/pull/15/commits/dd76b5c63dbd1116c65e172d4b8439d3af10ac3d

jmileham:
sweet.

PR #16, opened 2016-11-16
Add suggestion to run rake task for unknown splits

while1malloc0:
A common fix for local test track schema getting out of sync is to run `rake test_track:schema:load`. This PR puts that suggestion in the error messages for unknown splits.

### DIFF
--- a/app/models/test_track/ab_configuration.rb
+++ b/app/models/test_track/ab_configuration.rb
@@ -9,10 +9,7 @@ class TestTrack::ABConfiguration
 
     @true_variant = true_variant.to_s if true_variant
 
-    if @split_registry && !split
-      raise ArgumentError, "unknown split: #{split_name}. " \
-        "#{'You may need to run rake test_track:schema:load' unless Rails.env.production?}"
-    end
+    raise ArgumentError, unknown_split_error_message if @split_registry && !split
   end
 
   def variants
@@ -35,6 +32,12 @@ class TestTrack::ABConfiguration
   end
 
   attr_reader :split_name, :split_registry
+
+  def unknown_split_error_message
+    error_message = "unknown split: #{split_name}."
+    error_message << " You may need to run rake test_track:schema:load" if Rails.env.development?
+    error_message
+  end
 
   def split
     split_registry && split_registry[split_name]
--- a/app/models/test_track/ab_configuration.rb
+++ b/app/models/test_track/ab_configuration.rb
@@ -9,7 +9,10 @@ class TestTrack::ABConfiguration
 
     @true_variant = true_variant.to_s if true_variant
 
-    raise ArgumentError, "unknown split: #{split_name}. You may need to run 'rake test_track:schema:load'" if @split_registry && !split
+    if @split_registry && !split
+      raise ArgumentError, "unknown split: #{split_name}. " \
+        "#{'You may need to run rake test_track:schema:load' unless Rails.env.production?}"
+    end
   end
 
   def variants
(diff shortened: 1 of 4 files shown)


while1malloc0:
nanda

jmileham:
<< domain platform

As we talked about offline, we should probably only include this advice in `development` because it could overwrite admin-provided weightings in a prod-like environment. The kind of leakage that results in states like this seems to be churn and merge conflicts in feature branches of your consuming app, so the advice should be most useful in development, and not dangerous there.

tafn! Thanks @while1malloc0

while1malloc0:
bump

app/models/test_track/ab_configuration.rb
```diff
@@ -9,7 +9,10 @@ def initialize(opts)
 
     @true_variant = true_variant.to_s if true_variant
 
-    raise ArgumentError, "unknown split: #{split_name}" if @split_registry && !split
+    if @split_registry && !split
+      raise ArgumentError, "unknown split: #{split_name}. " \
+        "#{'You may need to run rake test_track:schema:load' unless Rails.env.production?}"
```
jmileham: I think this should be specific to development - in my experience most environments other than development are long-lived staging-type envs.

spec/models/test_track/ab_configuration_spec.rb
```diff
@@ -68,7 +68,7 @@
     it "raises a descriptive error when the split is not in the split_registry" do
       expect do
         described_class.new initialize_options.merge(split_name: :not_a_real_split)
-      end.to raise_error("unknown split: not_a_real_split")
+      end.to raise_error("unknown split: not_a_real_split. You may need to run rake test_track:schema:load")
```
jmileham: You kind of need to get into env overriding if you want to actually test this conditional display feature too, which I think is worth it because we don't want it to start popping up in production environments later.

jmileham:
tafn - lets get it constrained to development only and have tests for both cases. RSpec is very amenable to temporarily messing with the Rails.env value. If we don't have a safe `begin/ensure` block helper like we do for ENV vars for overriding Rails.env in tests it might be worth throwing one in.

app/models/test_track/ab_configuration.rb
```diff
@@ -9,7 +9,10 @@ def initialize(opts)
 
     @true_variant = true_variant.to_s if true_variant
 
-    raise ArgumentError, "unknown split: #{split_name}" if @split_registry && !split
+    if @split_registry && !split
+      raise ArgumentError, "unknown split: #{split_name}. " \
```
jmileham: Nit, but a little awkward to have a trailing space if you aren't showing more stuff. Put that in the conditional text I think?

spec/models/test_track/ab_configuration_spec.rb
```diff
@@ -68,7 +68,16 @@
     it "raises a descriptive error when the split is not in the split_registry" do
       expect do
         described_class.new initialize_options.merge(split_name: :not_a_real_split)
-      end.to raise_error("unknown split: not_a_real_split")
+      end.to raise_error("unknown split: not_a_real_split.")
+    end
```
jmileham: I'm interested in the production env behavior too - possibly more Than the test env behavior. One more spec demonstrating expected behavior in production and we should be good to go!

jmileham:
Come to think of it that might be a good reason to actually override the rails.env via Rails.env = instead of mocking.

while1malloc0:
bump @jmileham

jmileham:
great! domainlgtm platformlgtm!

jmileham:
I'm going to merge this and then cut a version @while1malloc0 - you can upgrade usages at your leisure. Thanks!

PR #17, opened 2016-11-22
TT connection errors and more

aburgel:
/domain @jmileham @ceslami @rynonl @samandmoore 

There are a couple of things in here.

1. Added some logging so we have a better idea how often we're getting these timeouts, cuz right now we just swallow them all.
2. Add `Faraday::ConnectionError` to the list of ignored exceptions, since these started happening more frequently with the switch to MRI (maybe jruby doesn't honor these timeouts?)
3. Attempt to make session management a little more lazy, so we don't attempt to load the assignments if we don't need to.

### DIFF
--- a/app/models/test_track/session.rb
+++ b/app/models/test_track/session.rb
@@ -129,7 +129,7 @@ class TestTrack::Session
   end
 
   def sync_assignments?
-    !visitor.offline? && visitor.unsynced_assignments.present?
+    visitor.loaded? && visitor.unsynced_assignments.present?
   end
 
   def mixpanel_distinct_id
--- a/app/models/test_track/remote/split_registry.rb
+++ b/app/models/test_track/remote/split_registry.rb
@@ -30,7 +30,8 @@ class TestTrack::Remote::SplitRegistry
         instance.attributes
       end.freeze
     end
-  rescue *TestTrack::SERVER_ERRORS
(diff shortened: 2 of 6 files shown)


jmileham:
Given that the rails client embeds the assignment hash in the web page for all ERB endpoints, is there a ton of value in being lazy about requesting the assignment bag? I suppose for API endpoints it could be.

app/models/test_track/session.rb
```diff
@@ -129,7 +129,7 @@ def create_alias!
   end
 
   def sync_assignments?
-    !visitor.offline? && visitor.unsynced_assignments.present?
+    visitor.loaded? && visitor.unsynced_assignments.present?
```
aburgel: Currently this call will trigger loading the remote assignments, if there were no earlier TT calls in this request. Since plenty of our requests don't have any splits in them, we're unnecessarily hitting TT.

I'm not sure I like exposing a `loaded?` method, it leaks too much of the internals of visitor into session. What about a method like `has_unsynced_assignments?` that will not trigger loading the remote assignments?

aburgel:
chatted offline with @jmileham: yes, even though rendered pages will trigger a load of assignments so they can be rendered on the page, this allows us to avoid having to tweak each base controller based on whether it expects to use those assignments.

aburgel:
bump

jmileham:
<< domain lgtm!

PR #18, opened 2016-12-07
Ability to sign_up new users from model mixin

rynonl:
/domain @jmileham @aburgel @samandmoore 

Currently the session discriminator works with either authenticated requests in a web context, or drops into an offline session that does not inspect the Test Track cookie at all. In practice, this is causing it to create a new visitor when we get a new signup and breaks the link between the new signup and the visitor he/she was prior to finishing the signup flow.

For brand new users, we want to `sign_up` to the Test Track server with their newly created application identifier and their existing visitor id that they've used up until this point. 

This PR adds a 3rd method that TestTrack::Identity injects into models it is included in called `test_track_sign_up`. It is meant to be called once per user signup and associates the local service's identifier(e.g. UserId) with the existing visitor id.  If an offline context is still inferred, we have no choice but to create a new visitor.

### DIFF
--- a/app/models/concerns/test_track/identity.rb
+++ b/app/models/concerns/test_track/identity.rb
@@ -46,6 +46,17 @@ module TestTrack::Identity
             end
           end
         end
+
+        define_method :test_track_sign_up! do
+          discriminator = TestTrack::IdentitySessionDiscriminator.new(self)
+          identifier_value = send(identifier_value_method)
+
+          if discriminator.participate_in_unauthenticated_session?
+            discriminator.controller.send(:test_track_session).sign_up! identifier_type, identifier_value
+          else
+            TestTrack::OfflineSession.create_visitor_for(identifier_type, identifier_value)
+          end
+        end
       end
     end
     # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
--- a/app/models/concerns/test_track/identity.rb
+++ b/app/models/concerns/test_track/identity.rb
@@ -11,7 +11,7 @@ module TestTrack::Identity
         define_method :test_track_ab do |*args|
           discriminator = TestTrack::IdentitySessionDiscriminator.new(self)
 
-          if discriminator.participate_in_online_session?
+          if discriminator.authenticated_resource_matches_identity?
             discriminator.controller.send(:test_track_visitor).ab(*args)
           else
             identifier_value = send(identifier_value_method)
@@ -24,7 +24,7 @@ module TestTrack::Identity
         define_method :test_track_vary do |*args, &block|
           discriminator = TestTrack::IdentitySessionDiscriminator.new(self)
 
-          if discriminator.participate_in_online_session?
+          if discriminator.authenticated_resource_matches_identity?
             discriminator.controller.send(:test_track_visitor).vary(*args, &block)
           else
             identifier_value = send(identifier_value_method)
@@ -37,7 +37,7 @@ module TestTrack::Identity
         define_method :test_track_visitor_id do
           discriminator = TestTrack::IdentitySessionDiscriminator.new(self)
 
-          if discriminator.participate_in_online_session?
+          if discriminator.authenticated_resource_matches_identity?
             discriminator.controller.send(:test_track_visitor).id
           else
             identifier_value = send(identifier_value_method)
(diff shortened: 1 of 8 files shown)


app/models/test_track/offline_session.rb
```diff
@@ -12,6 +12,10 @@ def self.with_visitor_for(identifier_type, identifier_value)
     end
   end
 
+  def self.create_visitor_for(identifier_type, identifier_value)
+    new(identifier_type, identifier_value).send :notify_unsynced_assignments!
```
jmileham: There should by definition be no unsynced assignments in this, right? because nothing happened during the session?

app/models/concerns/test_track/identity.rb
```diff
@@ -46,6 +46,17 @@ def test_track_identifier(identifier_type, identifier_value_method)
...
         end
+
+        define_method :test_track_sign_up! do
+          discriminator = TestTrack::IdentitySessionDiscriminator.new(self)
+          identifier_value = send(identifier_value_method)
+
+          if discriminator.participate_in_unauthenticated_session?
+            discriminator.controller.send(:test_track_session).sign_up! identifier_type, identifier_value
+          else
+            TestTrack::OfflineSession.create_visitor_for(identifier_type, identifier_value)
```
jmileham: I guess it doesn't do anything terrible, but it doesn't help much, because any later assignment done from the model layer will lazily create a visitor for this identifier and sign it up too. Maybe it should noop.

app/models/test_track/offline_session.rb
jmileham: Regardless, this method's behavior is probably moot because I think the caller either wants to blow up or noop instead of calling this method.

app/models/concerns/test_track/identity.rb
jmileham: Here's the code on the server side that a subsequent attempt to `with_visitor_for` would call: https://github.com/Betterment/test_track/blob/master/app/models/visitor_lookup.rb#L20-L27 which leads me to believe nooping here would be totally safe.

jmileham: And now to flip flop again, I'm thinking that if you call `test_track_sign_up!` you probably expect to be creating some kind of continuity with what happened before, which is impossible given that this can be credibly replaced with a noop if not participating in an unauthenticated session.  I think I'm back in the raise camp.

jmileham:
And I'd think that `test_track_log_in!` would also want to raise if there's no session to participate in, because it would likewise be a noop, and confusingly not result in creating any continuity either.

<< domain tafn

app/models/test_track/identity_session_discriminator.rb
```diff
@@ -13,6 +13,10 @@ def participate_in_online_session?
     authenticated_resource_matches_identity?
   end
 
+  def participate_in_unauthenticated_session?
+    web_context? && !controller.respond_to?(authenticated_resource_method_name, true)
```
aburgel: do these unauthenticated controllers not respond to `current_user` or does that method just return nil?

jmileham: Why would we care about whether it's an authenticated controller or not actually? It doesn't seem like if it's an authenticated controller we should go ahead and create a new visitor and throw it at the server?

jmileham: Maybe if we made the else side raise as I'm suggesting above you'd want to raise when trying to call sign_up from an authenticated controller, but I could imagine situations where auth frameworks actually do define and set that value on the controller. We should maybe stay out of it and just do our best to sign up if we're in a web_context?

app/models/concerns/test_track/identity.rb
rynonl: Yea I flip flopped all over this one also. The though was that if a new user was created via some job or as a side effect of another action we would want it to get a visitor and be trackable.

i.e.
```
# Non-web context, maybe a background job
user = User.new(name: 'foo')
user.do_trackable_thing

# Sometime later, actual person comes in to claim that user as there own, in web context
current_user = User.claim(name: 'foo')
current_user.do_other_trackable_thing
```

We'd want `do_trackable_thing` and `do_other_trackable_thing` to link back to the same visitor.  However, I think that we get that anyway per the visitor_lookup code you posted. `do_trackable_thing` would create a new visitor anyway.  

So yea, raise-y sgtm.

jmileham: yup! cool.

jmileham:
domain lgtm! Nice.

PR #20, opened 2016-12-09
appease rubocop

rynonl:

The default array indentation is pretty funky with the new rubocop version, so flipped it to be something closer to what we use elsewhere.

http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/IndentArray

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -81,6 +81,9 @@ Style/ClassAndModuleChildren:
 Style/AlignParameters:
   Enabled: false
 
+Style/IndentArray:
+  EnforcedStyle: consistent
+
 Style/ClosingParenthesisIndentation:
   Enabled: false
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -84,6 +84,9 @@ Style/AlignParameters:
 Style/IndentArray:
   EnforcedStyle: consistent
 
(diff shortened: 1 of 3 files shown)


jmileham:
Our rubocop.yml files should be in lockstep across repos - does this catch us up?

rynonl:
Doubtful, but I'm bumping versions in other projects as well and will update as necessary

rynonl:
bump

.rubocop.yml
```diff
@@ -81,6 +81,12 @@ Style/ClassAndModuleChildren:
 Style/AlignParameters:
   Enabled: false
 
+Style/IndentArray:
+  EnforcedStyle: consistent
```
samandmoore: 💯  i can't imagine why anyone would want anything else.

samandmoore:
> Our rubocop.yml files should be in lockstep across repos - does this catch us up?

@rynonl are we good on this?

rynonl:
@samandmoore 

We have ~10 rubocop files floating around.  They are roughly consistent, but not perfectly.  This change brings us to consistency with the changes I made to the configs in better-core...but this PR doesn't really encapsulate fixing allll the things that were already inconsistent.

So, yes?

samandmoore:
kk <<domainlgtm

PR #21, opened 2016-12-27
Raise on misconfiguration

aburgel:
/domain @jmileham @rynonl @samandmoore 

This does two things.

1) Fixes how we send events to Airbrake, it [expects an exception and an optional message](http://www.rubydoc.info/github/thoughtbot/airbrake/Airbrake.notify), not just a message. 

2) In development and test environments, raise when a `.vary` or `.ab` call is misconfigured.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,6 +15,7 @@ AllCops:
     - 'build/**/*'
   DisplayStyleGuide: true
   DisplayCopNames: true
+
 Rails:
   Enabled: true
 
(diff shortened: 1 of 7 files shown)


jmileham:
Interesting - I thought it worked this way before, but maybe it regressed at some point? Raising in development on misconfiguration was one of the key reasons to provide the explicit vary syntax.

jmileham:
Just the nokogiri compatibility thing and << domain lgtm!

aburgel:
It currently raise in some misconfiguration circumstances, but not all. I've updated the Gemfile, also had to update rubocop. I used our latest rubocop config.

PR #22, opened 2016-12-27
Revert raise on misconfig

aburgel:
/domain @jmileham @samandmoore @rynonl 

I'm reverting this because it didn't work with stubbed splits. When stubbing, we only register a single variant, so a vary call that refers to any other variant will get flagged as a misconfiguration.

There's no good way to know the actual variants of a split unless you load the `test_track_schema.yml` file, but that might not always be available in a test environment.

### DIFF
--- a/spec/assignment_helper_spec.rb
+++ b/spec/assignment_helper_spec.rb
@@ -23,5 +23,21 @@ RSpec.describe TestTrackRailsClient::AssignmentHelper do
         expect { stub_test_track_assignments(foo: :bar) }.to raise_error(/Cannot stub test track assignments/)
       end
     end
+
+    it 'works correctly with a vary call' do
+      stub_test_track_assignments(foo: :bar)
+
+      visitor = TestTrack::Visitor.new
+      expect do
+        visitor.vary(:foo, context: :spec) do |v|
+          v.when :bar do
+            # noop
+          end
+          v.default :baz do
+            raise "this branch shouldn't be executed"
+          end
+        end
+      end.not_to raise_error
+    end
   end
 end
--- a/spec/models/test_track/ab_configuration_spec.rb
+++ b/spec/models/test_track/ab_configuration_spec.rb
@@ -86,7 +86,7 @@ RSpec.describe TestTrack::ABConfiguration do
         with_rails_env 'production' do
           expect do
             described_class.new initialize_options.merge(split_name: :not_a_real_split)
-          end.not_to raise_error("unknown split: not_a_real_split. You may need to run rake test_track:schema:load")
+          end.to raise_error("unknown split: not_a_real_split.")
         end
       end
     end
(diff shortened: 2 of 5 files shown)


spec/assignment_helper_spec.rb
```diff
@@ -23,5 +23,21 @@
         expect { stub_test_track_assignments(foo: :bar) }.to raise_error(/Cannot stub test track assignments/)
       end
     end
+
+    it 'works correctly with a vary call' do
```
aburgel: I added a more integration-y test for `stub_test_track_assignments` that would have caught this issue sooner.

spec/models/test_track/ab_configuration_spec.rb
```diff
@@ -86,7 +86,7 @@
         with_rails_env 'production' do
           expect do
             described_class.new initialize_options.merge(split_name: :not_a_real_split)
-          end.not_to raise_error("unknown split: not_a_real_split. You may need to run rake test_track:schema:load")
+          end.to raise_error("unknown split: not_a_real_split.")
```
aburgel: The latest version of rspec didn't like this style: ```Using `expect { }.not_to raise_error(message)` risks false positives, since literally any other error would cause the expectation to pass```

jmileham:
Should we just do it for development and not test then?

samandmoore:
> There's no good way to know the actual variants of a split unless you load the test_track_schema.yml file, but that might not always be available in a test environment.

bump on john's suggestion to only raise in development. but also, when would there not be a  test_track_schema.yml in test environments?

jmileham:
I think it's that with the way stubbing works, awareness of the full split registry is hidden. Maybe we should move the specialized test mode behavior into the vary implementation where the full registry could be visible regardless of stubbing. That would also align with @rynonl's suggestion that the default behavior of splits in test mode not be pseudorandomized but instead go down the default path unless explicitly stubbed.

aburgel:
If the split lives in code coming from a Rails engine, the `test_track_schema.yml` may not be available. The schema file typically lives in a Rails app, so tests running in an engine, or an in another app that consumes the engine won't have access to the schema.

We could change that, but that would require changing some workflows, and possibly supporting multiple schemas per app?

aburgel:
I think we could follow @rynonl's suggestion without access to the registry. The `vary` call defines which variant is the default, so we could just use that information, and default to false for `ab` splits.

That might require a big conditional in the vary code, which doesn't feel right. I like the fact that we're mostly 'stubbing' data.

aburgel:
bump. changed to only raise in development

aburgel:
bump

spec/models/test_track/misconfiguration_notifier_spec.rb
```diff
@@ -5,27 +5,19 @@
 
   describe "#notify" do
     context "in development environment" do
-      around do |example|
+      it "raises" do
         with_rails_env "development" do
-          example.run
+          expect { subject.notify("something is misconfigured") }.to raise_exception 'something is misconfigured'
```
samandmoore: nit: just noticed we're using raise_error above but raise_exception here. not sure if they're different under the covers, but it seems like we only need to be using one of them.

samandmoore:
<<domainlgtm

PR #23, opened 2017-02-21
Use the requesting thread’s RequestStore when syncing assignments

aburgel:
/domain @rynonl @jmileham 

The notifier jobs call out to application code via the analytics plugin, but since we also fork a new thread when notifying unsynced assignments, we lose any per-request context in RequestStore.

This copies the calling thread's RequestStore into the notifying thread, so analytics plugins can still access that data.

### DIFF
--- a/app/models/test_track/session.rb
+++ b/app/models/test_track/session.rb
@@ -174,12 +174,12 @@ class TestTrack::Session
     ENV['TEST_TRACK_FULLY_QUALIFIED_COOKIE_DOMAIN_ENABLED'] == '1'
   end
 
-  def new_thread_with_request_store(&block)
+  def new_thread_with_request_store
     Thread.new(RequestStore.store) do |original_store|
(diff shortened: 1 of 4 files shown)


app/models/test_track/session.rb
```diff
@@ -114,8 +114,16 @@ def notify_unsynced_assignments!
       ##
       # This block creates an unbounded number of threads up to 1 per request.
       # This can potentially cause issues under high load, in which case we should move to a thread pool/work queue.
+      original_store = RequestStore.store
       Thread.new do
```
rynonl: You can do `Thread.new(RequestStore.store) do |original_store|` to make a local copy only in the child thread and avoid accessing any shared memory.

jmileham:
<< domain lgtm!

jmileham:
(assuming you can make rubocop happy)

PR #24, opened 2017-02-26
Stub split registry with default variants during testing

aburgel:
/domain @samandmoore @jmileham @rynonl 

This stubs the split registry before each test, setting one variant as the default so tests are deterministic. You'll no longer need to stub assignments before each spec, unless you have a particular need to.

I'm using the `test_track_schema.yml` file to generate the splits to stub. There are two issues with this.

1. The file will get loaded before each spec, which may slow things down a bit. We can do some caching, but I'm not sure where the best place to do this is.

2. The location of the file is fixed, which may not work in some apps. How should we make this configurable?

LMK if this approach seems sound and then I'll update the specs.

### DIFF
--- a/app/models/test_track/fake/visitor.rb
+++ b/app/models/test_track/fake/visitor.rb
@@ -1,5 +1,3 @@
-require 'digest'
-
 class TestTrack::Fake::Visitor
   attr_reader :id
 
@@ -21,21 +19,16 @@ class TestTrack::Fake::Visitor
     @assignments ||= _assignments
   end
 
+  def split_registry
+    TestTrack::Fake::SplitRegistry.instance.to_h
+  end
+
   private
 
   def _assignments
-    TestTrack::Fake::SplitRegistry.instance.splits.map do |split|
-      index = hash_fixnum(split.name) % split.registry.keys.size
-      variant = split.registry.keys[index]
-      Assignment.new(split.name, variant, false, "the_context")
+    split_registry.keys.map do |split_name|
+      variant = TestTrack::VariantCalculator.new(visitor: self, split_name: split_name).variant
+      Assignment.new(split_name, variant, false, "the_context")
     end
-  end
-
-  def hash_fixnum(split_name)
-    split_visitor_hash(split_name).slice(0, 8).to_i(16)
-  end
-
-  def split_visitor_hash(split_name)
-    Digest::MD5.new.update(split_name.to_s + id.to_s).hexdigest
   end
 end
--- a/lib/test_track_rails_client/assignment_helper.rb
+++ b/lib/test_track_rails_client/assignment_helper.rb
@@ -1,13 +1,12 @@
 module TestTrackRailsClient::AssignmentHelper
-  def stub_test_track_assignments(assignment_registry) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+  def stub_test_track_assignments(assignment_registry) # rubocop:disable Metrics/AbcSize
     raise "Cannot stub test track assignments when TestTrack is enabled" if TestTrack.enabled?
 
-    split_registry = {}
+    split_registry = TestTrack::Fake::SplitRegistry.instance.to_h.dup
     assignments = []
 
     assignment_registry.each do |split_name, variant|
-      assignment_registry[split_name] = variant.to_s
-      split_registry[split_name] = { variant => 100 }
+      split_registry[split_name] = { variant => 100 } unless split_registry[split_name]
       assignments << { split_name: split_name.to_s, variant: variant.to_s, unsynced: false }
     end
 
--- a/app/models/test_track/fake/split_registry.rb
+++ b/app/models/test_track/fake/split_registry.rb
@@ -6,11 +6,7 @@ class TestTrack::Fake::SplitRegistry
   end
 
   def to_h
-    if test_track_schema_yml.present?
-      test_track_schema_yml[:splits]
-    else
-      {}
-    end
+    @to_h ||= splits_with_deterministic_weights
   end
 
   def splits
@@ -21,6 +17,14 @@ class TestTrack::Fake::SplitRegistry
 
   private
 
+  def split_hash
+    if test_track_schema_yml.present?
+      test_track_schema_yml[:splits]
+    else
+      {}
+    end
+  end
+
   def test_track_schema_yml
     unless instance_variable_defined?(:@test_track_schema_yml)
       @test_track_schema_yml = _test_track_schema_yml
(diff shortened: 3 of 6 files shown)


lib/test_track_rails_client/assignment_helper.rb
```diff
@@ -1,19 +1,39 @@
...
-    split_registry = {}
-    assignments = []
+    schema = TestTrackRailsClient::Schema.new
 
+    assignments = []
     assignment_registry.each do |split_name, variant|
-      assignment_registry[split_name] = variant.to_s
-      split_registry[split_name] = { variant => 100 }
+      raise "Variant #{variant} for #{split_name} must be a string or symbol" unless [String, Symbol].include?(variant.class)
+      raise "Split #{split_name} does not exist in #{schema.file_path}" unless schema.splits.include?(split_name)
```
aburgel: Now that we have access to the schema, we can validate that the splits we're stubbing actually exists.

samandmoore:
Oh wow! We were going to work on this in eng efficiency next week https://app.asana.com/0/233142518428541/279889518446809

There are some details in the task that might clarify some of your questions. Not sure. I will take a closer look later tonight.

samandmoore:
On point 2, that's not a new thing right? Wasn't the schema file always fixed?

jmileham:
The code that does the fuzzing is still somewhere in there and I don't see it in this diff either. The variants are actually deterministic already over the same rspec seed, but we were talking about pulling that in favor of simpler determinism (alpha first). I assume that asana card captures all the deets.

lib/test_track_rails_client/assignment_helper.rb
```diff
@@ -1,19 +1,39 @@
...
+      default_variant = weighting_registry.keys.sort.first
+
+      adjusted_weightings = { default_variant => 100 }
+      weighting_registry.except(default_variant).keys.each do |variant|
+        adjusted_weightings[variant] = 0
+      end
+
+      split_registry[split_name] = adjusted_weightings
+    end
+
```
samandmoore: So, I think going along with what John was saying, we may be able to push this code down into the place where we are currently doing the fuzzing as a replacement approach. Then we will also get this behavior in development if someone isn't relying on a real TT server.

aburgel:
Oops! Didn't realize you were working on this already. There were no good movies on my flight so I was looking for something I could get done in a couple of hours.

Double oops, didn't know about that faking thing either. So much for hacking something together real quick 😜

I'll update this in the next couple of hours.

aburgel:
Also, my point #2 wasn't clear. I meant that if you have split code in an engine that is shared by 2 apps, the scheme file will only live with one of the apps. The second app won't have access to that yaml file so we'll need to make the path configurable or something along those lines.

samandmoore:
Ah yea. We can use an ENV variable. That's a good idea in general.

jmileham:
There is also the tension about being able to stub a split that comes from another app, which is legal. So whatever we do, unless we change the legality of that, we probably want to allow stubbing of unknown splits. Haven't read carefully enough to know whether that's a thing. Disallowing stubbing of unknown variants of known splits does seem legit though.

lib/test_track_rails_client/rspec_helpers.rb
```diff
@@ -2,4 +2,8 @@
 
 RSpec.configure do |config|
   config.include TestTrackRailsClient::AssignmentHelper
+
+  config.before(:each) do
```
rynonl: regarding loading the schema file before each spec, seems that we could do something like make Schema a singleton that loads the file once, this line becomes a `config.before(:all)`, `stub_test_track_assignments` mutates an in memory registry, and in a `config.before(:each)` we reset it back to the default.

samandmoore: This might not be necessary at all if we fix the fuzzing stuff. I'm pretty sure that the fuzzing logic needs to read the yml file already. I would wait on making any changes to stuff at the rspec level until we figure out if we need it at all.

aburgel:
bump. i moved the new deterministic split registry logic into the fake split registry and set a before hook to stub out the local split registry to use the fake split registry as its backing store. the nice thing about this is that now the js client will see the same split registry as the ruby code when running feature specs.

i'll annotate a few other interesting changes. and i still need to add a test or two as well.

app/models/test_track/fake/visitor.rb
```diff
@@ -21,21 +19,16 @@ def assignments
...
+
   private
 
   def _assignments
-    TestTrack::Fake::SplitRegistry.instance.splits.map do |split|
-      index = hash_fixnum(split.name) % split.registry.keys.size
-      variant = split.registry.keys[index]
-      Assignment.new(split.name, variant, false, "the_context")
+    split_registry.keys.map do |split_name|
+      variant = TestTrack::VariantCalculator.new(visitor: self, split_name: split_name).variant
```
aburgel: I changed this to use the VariantCalculator rather than the partial implementation we had before. I think it cleans things up a bit, but its not strictly necessary.

lib/test_track_rails_client/rspec_helpers.rb
```diff
@@ -2,4 +2,8 @@
...
+    TestTrack::FakeServer.reset!(RSpec.configuration.seed)
```
samandmoore: do we need this anymore? we're using a stub in the helper above which will be automatically undone at the end of each spec. and the split registry is sourced from the schema file and shouldn't be mutated at all during test runs, right?

aburgel: You do need to mutate the split registry if you're stubbing a split that doesn't exist in the registry, otherwise the vary call will blow up saying that the split isn't in the registry.

aburgel: I suppose we could continue to stub the remote split registry here: https://github.com/Betterment/test_track_rails_client/pull/24/files#diff-eca6a7243d1a365a9354cf12853389bcL17.

But then the ruby and js clients would see different split registries.

app/models/test_track/fake/visitor.rb
aburgel: Actually, scratch that. It is necessary because previously we were just picking any variant. Now we need to pick the correct one based on the variant weights.

samandmoore: yea, i think this is good. no need to unwrap that logic into here.

lib/test_track_rails_client/rspec_helpers.rb
samandmoore: hmm. yea, i think i like stubbing that registry there. having this here in order to clean up after that stub method is very unclear to me.

samandmoore:
<<domainlgtm! this looks great.

PR #26, opened 2017-04-05
Split Detail Remote Model

agirlnamedsophia:
/platform @dschaub @samandmoore 

* This adds a SplitDetail remote model that hits the `split_details` endpoint in test_track 
* relies on [this PR](https://github.com/Betterment/test_track/pull/43/files) to be merged first


app/models/test_track/remote/split_detail.rb
```diff
@@ -0,0 +1,30 @@
+class TestTrack::Remote::SplitDetail
+  include TestTrack::RemoteModel
+
+  collection_path '/api/v1/split_details'
+
+  attributes :name
+
+  def self.from_name(name)
+    raise "must provide a name" unless name.present?
```
dschaub: is this necessary? if you pass a blank value you'll get a 404, and if you leave the argument out you'll get a ruby error

app/models/test_track/remote/split_detail.rb
```diff
@@ -0,0 +1,30 @@
...
+    if faked?
+      new(fake_instance_attributes(nil))
+    else
+      get("/api/v1/split_details/#{name}")
+    end
+  end
+
+  def self.fake_instance_attributes(_)
+    {
+      split_name: "fake_visitor_id",
```
dschaub: fake_split_name?

app/models/test_track/remote/split_detail.rb
```diff
@@ -0,0 +1,30 @@
...
+      hypothesis: "fake_hypothesis",
+      assignment_criteria: "fake_criteria",
+      description: "fake_description",
+      owner: "fake_retail",
+      location: "fake_activity",
+      platform: "fake_mobile"
```
dschaub: that should be mobile cause it can only be one of two values, right?

spec/models/test_track/remote/split_detail_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    subject { described_class.find("fake_split_name_from_server") }
+
+    it "loads split details with fake instance attributes" do
+      expect(subject.split_name).to eq("fake_visitor_id")
+      expect(subject.hypothesis).to eq("fake_hypothesis")
+      expect(subject.assignment_criteria).to eq("fake_criteria")
+      expect(subject.description).to eq("fake_description")
+      expect(subject.owner).to eq("fake_retail")
+      expect(subject.location).to eq("fake_activity")
+      expect(subject.platform).to eq("fake_mobile")
```
dschaub: s/fake_mobile/mobile/

dschaub:
<< domain tafn

agirlnamedsophia:
nanda?

agirlnamedsophia:
accidental clik :(

app/models/test_track/remote/split_detail.rb
```diff
@@ -0,0 +1,28 @@
+class TestTrack::Remote::SplitDetail
+  include TestTrack::RemoteModel
+
+  collection_path '/api/v1/split_details'
+
+  attributes :name
```
dschaub: missed this, i think we want all of the other attributes here too

dschaub:
domain lgtm

app/models/test_track/fake/schema_loading.rb
```diff
@@ -0,0 +1,34 @@
+class TestTrack::Fake::SchemaLoading
```
dschaub: we could probably just name this `Schema`

app/models/test_track/fake/schema_loading.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_reader :split_hash
+
+  def initialize
+    @split_hash = _split_hash
+  end
```
dschaub: can we make this a public method called `split_hash` that memoizes the result of `_split_hash`? (or uses the `begin...end` style

app/models/test_track/fake/schema_loading.rb
dschaub: also, worth it to make this a singleton? otherwise if we have multiple uses it'll load the same schema file twice, right?

app/models/test_track/fake/split.rb
```diff
@@ -0,0 +1,20 @@
+class TestTrack::Fake::Split
+  def self.instance
+    @instance ||= new
+  end
```
dschaub: can we just use `Singleton` for this?

spec/controllers/tt/api/v1/split_details_controller_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe Tt::Api::V1::SplitDetailsController do
+  let(:response_json) { JSON.parse(response.body) }
+
+  describe '#show' do
+    it 'returns fake split details' do
+      get :show, format: :json
+      expect(response_json).to eq TestTrack::FakeServer.split_details
```
dschaub: do we want to list out the expected values? this seems like it's just asserting that a variable is equal to itself

agirlnamedsophia:
nanda?

spec/dummy/db/test_track_schema.yml
```diff
@@ -1,9 +1,23 @@
...
-    'true': 50
+    weights:
+      'false': 50
+      'true': 50
+    hypothesis: "user will buy more hats with promotion"
+    location: "hat store"
+    platform: "desktop"
+    owner: "hat makers"
+    assignment_criteria: "user has bought one hat"
+    description: "promotion test to see if users will buy more hats"
```
samandmoore: is this actually getting dumped into the schema.yml now? i don't think that it is and i don't think that it should be.

the schema.yml is only meant to track the critical details of splits, their variants, and the associated weighting. we don't want it to include any of the optional split detail data that is inputted through the admin ui.

app/models/test_track/fake/split.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def split_details
+    @split_details ||= _split_details
+  end
+
+  private
+
+  def _split_details
+    split_hash = TestTrack::Fake::Schema.instance.split_hash
+    { "name" => "banner_color" }.merge(split_hash["banner_color"].except("weights"))
```
samandmoore: going along with the comment i left elsewhere about not putting the split details in the schema.yml file, i think that we should just have this return fake data just like we do in the `fake_instance_attributes` method.

i am also noticing that this fake doesn't include any of the details about the variants in this split. is that an intentional omission?

app/models/test_track/fake/split.rb
```diff
@@ -0,0 +1,14 @@
+class TestTrack::Fake::Split
```
samandmoore: i think this is actually SplitDetail, right?

agirlnamedsophia:
nanda?

agirlnamedsophia:
alright, I think I've got it!

nanda?

dschaub:
domain lgtm

samandmoore:
domainlgtm

samandmoore:
nanda?

samandmoore:
<<domainlgtm again

samandmoore:
oh. i can't read. sigh.

<<platformlgtm

PR #27, opened 2017-04-11
visitor details model and fakes

dschaub:
/domain @Betterment/test_track_core @agirlnamedsophia 

fake data for visitor details, which are the human readable forms of a visitor's assignments


app/controllers/tt/api/v1/visitor_details_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Tt::Api::V1::VisitorDetailsController < Tt::Api::V1::ApplicationController
+  def show
+    render json: TestTrack::FakeServer.visitor_details
```
dschaub: I noticed that we're doing more complicated stuff to model out this fake data for other models, when really it seems like we just want to dump out a pre-formed json response. am I thinking about that wrong? this seemed like the simplest way to achieve that.

samandmoore: Yea, I _think_ this is fine. We really just want to make this fakes as lifelike as they need to be in order to work. They can be made more and more lifelike as we have a stronger need for that.

So, I think this is fine.

dschaub: I was thinking this morning that this still has a bunch of unnecessary indirection, when really we could just have a `show.json`. but I guess it makes sense to keep it consistent with the others and centralize around `FakeServer`.

samandmoore: Yea, the idea is that these things could become stateful over time, if we find the need. So, centralizing it is nice. Also, FakeServer has a seed fed into it so it's good that we have them all hanging off of the central object so that we can easily leverage the seed or other injected bits if we so desire.

samandmoore:
<<domain

It might be useful to include more than one item in assignment_details, but I don't have strong feelings.

lgtm

samandmoore:
domainlgtm -- whoops on the lack of dmain on the last one.

PR #28, opened 2017-04-11
Display variant details

agirlnamedsophia:
* This includes variant details in the fake data

/domain @Betterment/test_track_core @dschaub @samandmoore

### DIFF
--- a/spec/models/test_track/remote/split_detail_spec.rb
+++ b/spec/models/test_track/remote/split_detail_spec.rb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe TestTrack::Remote::SplitDetail do
-  let(:fake_details) {
+  let(:fake_details) do
     {
       name: "great_name",
       hypothesis: "fake hypothesis",
@@ -21,7 +21,7 @@ RSpec.describe TestTrack::Remote::SplitDetail do
         }
       ]
     }
-  }
+  end
 
   before do
     stub_request(:get, url)
--- a/spec/models/test_track/remote/split_detail_spec.rb
+++ b/spec/models/test_track/remote/split_detail_spec.rb
@@ -1,6 +1,28 @@
 require 'rails_helper'
 
 RSpec.describe TestTrack::Remote::SplitDetail do
+  let(:fake_details) {
+    {
+      name: "great_name",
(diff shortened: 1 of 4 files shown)


spec/models/test_track/remote/split_detail_spec.rb
```diff
@@ -27,6 +37,12 @@
       expect(subject.owner).to eq("fake owner")
       expect(subject.location).to eq("fake activity")
       expect(subject.platform).to eq("mobile")
+
+      expect(subject.variant_details.size).to eq 2
+      expect(subject.variant_details.first[:name]).to eq("fake first variant detail")
+      expect(subject.variant_details.first[:description]).to eq("There are FAQ links in a sidebar")
+      expect(subject.variant_details.last[:name]).to eq("fake second variant detail")
+      expect(subject.variant_details.last[:description]).to eq("There are FAQ links in the default footer")
```
dschaub: we generally want to assert against things that are defined in the test file somewhere, so usually we'd mock the method that returns these values to return something else.

agirlnamedsophia: that happens below in the `with_test_track_enabled` block. 

these values are the default `fake_instance_attributes` from the Her model.

dschaub: right, I mean we want to be mocking the `fake_instance_attributes` method in here to return the values that we want to assert against, instead of asserting against values that live in a file elsewhere

agirlnamedsophia: ahhh I see, I see. 👍

agirlnamedsophia:
nanda?

spec/models/test_track/remote/split_detail_spec.rb
```diff
@@ -1,6 +1,28 @@
...
+          name: "fake first variant detail",
+          description: "There are FAQ links in a sidebar"
+        },
+        {
+          name: "fake second variant detail",
+          description: "There are FAQ links in the default footer"
+        }
+      ]
+    }
+  }
```
dschaub: rubocop will probably make you change this to a `do...end`

dschaub:
just that block style in the test, << domain lgtm

PR #29, opened 2017-04-28
Allow test_track_visitor_id to be passed via headers

betterzega:
/domain @jmileham @samandmoore @aburgel

From our conversation in the test_track channel, allowing the session to be read from the request header.

Still need to write back the test track visitor id in the response header. Should this only be passed back when the request has the TT visitor id in the header, or should it always be passed back?

Still need to do a proper integration test.

### DIFF
--- a/app/models/test_track/visitor.rb
+++ b/app/models/test_track/visitor.rb
@@ -98,6 +98,10 @@ class TestTrack::Visitor
     !offline? && @remote_visitor.present?
   end
 
+  def id_overridden_by_existing_visitor?
+    @id_overridden_by_existing_visitor || false
+  end
+
   private
 
   def assignments
@@ -113,6 +117,7 @@ class TestTrack::Visitor
   end
 
   def merge!(other)
+    @id_overridden_by_existing_visitor = id != other.id
     @id = other.id
     @assignment_registry = assignment_registry.merge(other.assignment_registry)
     @unsynced_assignments = nil
--- a/spec/models/test_track/session_spec.rb
+++ b/spec/models/test_track/session_spec.rb
@@ -1,10 +1,12 @@
 require 'rails_helper'
 
 RSpec.describe TestTrack::Session do
-  let(:controller) { instance_double(ApplicationController, cookies: cookies, request: request) }
+  let(:controller) { instance_double(ApplicationController, cookies: cookies, request: request, response: response) }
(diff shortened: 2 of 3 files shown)


jmileham:
> Should this only be passed back when the request has the TT visitor id in the header, or should it always be passed back?

I think it should only be passed back when it changes, probably? Like a `X-Set-TT-Visitor-ID`. It would only be set when you call `test_track_visitor.log_in!` or `test_track_visitor.sign_up!`. We might need to plumb the controller ref deeper into that DSL in order to make that work, but that seems like the right move to me.

app/models/test_track/visitor.rb
```diff
@@ -98,6 +98,10 @@ def loaded?
     !offline? && @remote_visitor.present?
   end
 
+  def id_changed?
```
betterzega: Tried to follow rails naming conventions, but maybe this hides what this actually represents, the fact that the visitor id was overridden by the remote visitor?

betterzega:
bump

app/models/test_track/visitor.rb
jmileham: Yeah, I think you might want to go with a special name here. Remote won't have as much meaning to a browser of the controller mixin because to them the remote in mind is the client of the controller, not the upstream TT server, so some name that captures that the TT server changed our identity would be hot.

spec/models/test_track/session_spec.rb
```diff
@@ -59,6 +61,38 @@
...
+          identifier = instance_double(TestTrack::Remote::Identifier, visitor: real_visitor)
+          allow(TestTrack::Remote::Identifier).to receive(:create!).and_return(identifier)
+        end
+
+        it "does not return the visitor id in the reponse header" do
+          subject.manage do
+            subject.log_in!("identifier_type", "value")
+          end
+
+          expect(controller.response.headers['X-Set-TT-Visitor-ID']).to eq(nil)
```
jmileham: not_to have_key or something?

jmileham:
<< domain tafn. Just come up with an awesome name and this is looking great.

app/models/test_track/visitor.rb
betterzega: `id_overridden_by_existing_visitor?`?

jmileham:
that'll work. it's a long enough name that it doesn't give you unwarranted confidence that you know what it's doing. domain lgtm!

PR #31, opened 2017-06-13
Explicitly show possible url for TEST_TRACK_API_URL

### LINKED ISSUE #30: Explicitly show the TEST_TRACK_API_URL domain
The document only shows the .dev domain which works only for pow.cx user. For other users, we need to use `localhost:PORT_NUMBER` instead.

Change from `http://[myapp]:[your new app password]@testtrack.dev/` to `http://[myapp]:[your new app password]@[your-app-domain]/` or adding a note would be great.

vtno:
Fix #30


jmileham:
Looks great. Thank you! Looks like the build broke for completely unrelated reasons. I'll get that patched up and merge away. Thanks for your contribution!

PR #32, opened 2017-07-25
Test against multiple Rails versions

smudge:
/platform @samandmoore @jmileham 

Using the `appraisal` gem to provide Travis with multiple Gemfile variations.

I excluded combinations of Ruby and Rails that are not supposed to be compatible. There is also a [jruby DB adapter issue](https://github.com/jruby/activerecord-jdbc-adapter/issues/780) preventing things from working on Rails 5.1, so I have excluded that combination as well.

I also discovered an issue with JRuby < 9.1.9.0 and Rails >= 4.2.9, which was causing the tests to fail because `Rails.cache.fetch` was always missing. For some reason it was not an issue with Rails <= 4.2.8, so this may be worth further investigation for anyone using this gem on JRuby.


.travis.yml
```diff
@@ -2,12 +2,36 @@ language: ruby
 
 rvm:
 - 1.9.3
-- jruby-19mode
+- jruby-9.1.9.0-d19
```
samandmoore: d19? does that run in 19mode or something?

samandmoore:
<<domainlgtm platformlgtm

.travis.yml
smudge: Yeah, it's 19mode but on a newer engine. (To avoid that issue I mentioned above, which affected both jruby modes)

PR #33, opened 2017-07-26
Establish OfflineSession with a test_track_visitor_id

jszymanowski:
/domain @Betterment/test_track_core @ceslami 
/platform @samandmoore @jmileham @aburgel @ceslami 

Allow for instantiation of an `OfflineSession` when a `test_track_visitor_id` is available in lieu of `identifier_type`/`identifier_id`.


jszymanowski:
retest this please

aburgel:
Should we run this through the Identity mixin or something similar?

I think the changes to the OfflineSession are correct, but there's an optimization that we can make since in some cases, we may be able to reuse the current web session. Its likely that the web session is operating on the same tt visitor id.

We basically need to follow the same discriminator pattern, and check whether the web session tt visitor id matches the one we're making the request on behalf of.

ceslami:
Aha! Chatted offline a bit, and it is indeed an interesting question worth asking.

The ask is, most simply: how can we most-universally take advantage of an online session (ie discrimination), if one exists?

In our use-case, the only thread we have is a `TestTrack::Visitor#id`. In order to get a session, we need  the option to manifest an `OfflineSession`, so that our use-case isn't coupled to online contexts. The idea of using discrimination to reuse an online session, if one exists, is appealing. That said, we don't have a `TestTrack::Identity`-able entity at-hand.

I think this work paves the way for -- and begs the question of -- whether `TestTrack::Visitor`, or a neighboring entity, can have `TestTrack::Identity`. Right now, an `identifier_type` is a requirement of a `TestTrack::Identity`, and `:test_track_visitor` is not one-such `identifier_type`. So there are some interesting questions to answer before we get there.

Given that this work would be necessary for such an enhancement, I'd love to get it merged, and kick off a design discussion about how we want `TestTrack::Identity`, for the purposes of leveraging discrimination, to work when a `TestTrack::Visitor#id` is our only handle.

aburgel:
<< platform lgtm

README.md
```diff
@@ -248,6 +251,23 @@ OfflineSession.with_visitor_for(:myapp_user_id, 1234) do |test_track_visitor|
...
+    end
+    v.when :variant_3 do
+      # Do another thing
+    end
+    v.default :variant_4 do
+      # Do something else
+    end
+  end
+end
+```
```
ceslami: Nice!

README.md
```diff
@@ -248,6 +251,23 @@ OfflineSession.with_visitor_for(:myapp_user_id, 1234) do |test_track_visitor|
 end
 ```
 
+2. with a TestTrack `visitor_id`:
```
ceslami: We might want to be clearer about referring to this identifier. `TestTrack::Visitor#id` is how I'd stylize it, but we also use "visitor ID" elsewhere in the docs. I might opt for the former for clarity.

ceslami:
<< domain tafn

README.md
jszymanowski: Sounds good to me.

jszymanowski:
bump

ceslami:
<< domain lgtm!

PR #35, opened 2017-09-05
Use main session after signup/login

aburgel:
/domain @samandmoore @jmileham @ceslami 
/no-platform
/cc @joejansen @chrislopresto 

This addresses a performance problem in sign up requests (and for logins too). After calling `test_track_sign_up!`, subsequent `vary` calls on the signed up identity will use an offline session, even though the main session could be used (since it's visitor now has that identity). Using an offline session means at least one, possibly two, additional remote calls per `vary`.

This changeset modifies `TestTrack::Session` to keep track of signed up and logged in identities, so the session discriminator can use the main session if its operating on one of those identities. I also cleaned up the Identity mixin a bit, so it delegates more to the discriminator.

This is a breaking change on `TT::Session`, so we'll have to bump the major version number. I could make this backwards compatible, but the upgrade path is pretty straightforward.

TODO
- [x] Update specs
- [x] Add an `UPGRADING` section to the readme

### DIFF
--- a/app/models/test_track/session.rb
+++ b/app/models/test_track/session.rb
@@ -36,7 +36,7 @@ class TestTrack::Session
 
     @visitor = TestTrack::Visitor.new if opts[:forget_current_visitor]
     visitor.link_identifier!(identity.test_track_identifier_type, identity.test_track_identifier_value)
-    managed_identities << identity
+    matching_identities << identity
     self.mixpanel_distinct_id = visitor.id
     true
   end
@@ -45,13 +45,12 @@ class TestTrack::Session
     raise "must provide a TestTrack::Identity" unless identity.is_a?(TestTrack::Identity)
 
     visitor.link_identifier!(identity.test_track_identifier_type, identity.test_track_identifier_value)
-    managed_identities << identity
+    matching_identities << identity
     @signed_up = true
   end
 
-  def managed_identity?(identity)
-    managed_identity = managed_identities.find_by_identifier_type(identity)
-    managed_identity.present? && managed_identity == identity
+  def has_matching_identity?(identity)
+    matching_identities.include?(identity)
   end
 
   private
@@ -227,7 +226,7 @@ class TestTrack::Session
     end
   end
 
-  def managed_identities
-    @managed_identities ||= TestTrack::SessionIdentityCollection.new(controller)
+  def matching_identities
+    @matching_identities ||= TestTrack::SessionIdentityCollection.new(controller)
   end
 end
--- a/app/models/test_track/session.rb
+++ b/app/models/test_track/session.rb
@@ -47,7 +47,7 @@ class TestTrack::Session
     @visitor = TestTrack::Visitor.new if opts[:forget_current_visitor]
     visitor.link_identifier!(identifier_type, identifier_value)
 
-    matching_identities << identity if identity.present?
+    identities << identity if identity.present?
     self.mixpanel_distinct_id = visitor.id
     true
   end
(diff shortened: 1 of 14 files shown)


app/models/test_track/session.rb
```diff
@@ -31,23 +31,41 @@ def state_hash
     }
   end
 
-  def log_in!(identifier_type, identifier_value, opts = {})
+  def log_in!(identity, opts = {})
```
aburgel: This is the breaking change, but I think its a much nicer contract:

`session.log_in!('user_id', user.id)` vs `session.log_in!(user)`

Of course, this requires that User mixes in `TT::Identity`.

jmileham: and we're relying on conformance to the duck type of responding to `test_track_identifier_type` and `test_track_identifier_value` as our verification of mixing in `TT::Identity`? That would mean that a mistake would result in a method missing error rather, right? Should we have it check for inheritance from `TT::Identity` and throw a more actionable error message?

aburgel: I was debating adding a check. I left it out because I thought relying on duck typing was more ruby-esque. I can add it tho. For this new feature to work, you do need to pass in something that inherits from `TT::Identity`, so we might as well require it.

app/models/test_track/session.rb
```diff
@@ -31,23 +31,41 @@ def state_hash
...
-    visitor.link_identifier!(identifier_type, identifier_value)
+  def sign_up!(identity)
+    visitor.link_identifier!(identity.test_track_identifier_type, identity.test_track_identifier_value)
+    managed_identities[identity.test_track_identifier_type] = identity
     @signed_up = true
   end
 
+  def managed_identity?(identity)
+    managed_identity = managed_identities[identity.test_track_identifier_type] ||= authenticated_resource_for_identity(identity)
+    managed_identity.present? && managed_identity == identity
```
jmileham: a lot going on in these two lines. This feels a little bit like object envy maybe? Like this might be easier to grok if `IdentityCollection`were another class or something? like:

```ruby
class TestTrack::Session
  ...

  def managed_identities
    @managed_identities = SessionIdentityCollection.new(controller)
  end

  ...
end

class SessionIdentityCollection
  def initialize(controller)
    ...
  end

  def find_by_identifier_type(identifier_type)
    ...
  end

  def <<(identity)
    ...
  end
end
```

jmileham: Is `managed_identity?` only going to return false if the controller-authenticated identity doesn't match the requested identity? Maybe we should name the methods for getting at this fact to reflect that authorization concern? Because if I'm reading correctly it'll lazily start managing any identity not in conflict with the controller identity and return true, so `managed` doesn't sound right?

jmileham: `Session#has_matching_identity?(identity)` or something?

jmileham:
<< domain tafn

app/models/test_track/session.rb
samandmoore: I think it would still be sufficiently rubyish to check if it `responds_to?` the required methods or raise.

aburgel:
@jmileham @samandmoore bump

aburgel:
> Should we allow a temporary warning-inducing acceptance of the previous signature that works ok without dirtying up our internal implementation?

I could go either way on this. My intention was to bump the version to 2.0, which should be a signal that there's a breaking change. I don't have a sense for how big the community and how well they test their code.

Seems easy enough to make this backwards compatible, so I'm happy to do it if you feel strongly. @samandmoore thoughts?

samandmoore:
as i see it, our community is fairly small, but if the cost of making this backward-compatible is "easy enough" then i'd definitely vote to do it.

aburgel:
bump. made it backwards compatible with a warning.

i kept the upgrade notes in the readme, but changed the upcoming version to 1.3 since I think we want a minor version bump.

jmileham:
<< domain lgtm!

PR #36, opened 2017-10-10
Loosen public suffix constraint

while1malloc0:
/platform @samandmoore @jmileham 

TestTrack's public_suffix dependency is outdated, and the version constraint on it is more restrictive than it probably needs to be. 

The largest behavior change in public_suffix 2+ is that lookups now default to a "wildcard" rule, for validations, which is much more permissive in the domains that it allows. The old behavior is preserved here by using the `default_rule: nil` argument to `PublicSuffix.parse`. 

Upgrading public suffix also requires dropping compatibility for ruby versions < 2.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 
 rvm:
-- 1.9.3
 - jruby-9.1.9.0-d19
 - 2.0.0
 - 2.3.1
--- a/app/models/test_track/session.rb
+++ b/app/models/test_track/session.rb
@@ -123,7 +123,7 @@ class TestTrack::Session
   end
 
   def public_suffix_host
-    @public_suffix_host ||= PublicSuffix.parse(request.host)
+    @public_suffix_host ||= PublicSuffix.parse(request.host, default_rule: nil)
   end
 
   def manage_cookies!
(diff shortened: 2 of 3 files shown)


samandmoore:
<<domainlgtm
<<platformlgtm

it's worth noting that we're also raising the min-version of publicsuffix as well.

samandmoore:
it looks like we broke ruby 1.9.3 compat with this update. we should circle up offline to make an active decision about that change.

samandmoore:
/domain @samandmoore
<<domain tafn -- let's discuss the 1.9 compat thing

aburgel:
I think we can remove PublicSuffix completely. Its only used in two places that I think can be handled without it.

1. [`fully_qualified_domain`](https://github.com/Betterment/test_track_rails_client/blob/master/app/models/test_track/session.rb#L117) can be replaced by `request.host`. I don't think PublicSuffix adds much except rejecting bad domains, which doesn't seem like a problem that TestTrack needs to solve.

2. [`wildcard_domain`](https://github.com/Betterment/test_track_rails_client/blob/master/app/models/test_track/session.rb#L121) can be replaced with [`:all`](http://api.rubyonrails.org/v4.2/classes/ActionDispatch/Cookies.html), which is a Rails-provided solution to this problem.

aburgel:
Hmm, maybe not. We also stick the cookie domain in state hash which is read by the TT JS client. So we need the actual cookie domain, not `:all`. Rails handles that internally and we don't get to see the result.

samandmoore:
so, i think we can just keep the library in here then, right?

samandmoore:
LGTM

i will merge and release as 2.0 with support for ruby 1.9.3 dropped.

Issue #8, opened 2016-11-01
#vary allows specifying a default that is already handled


rynonl:
Ex:

```
.vary :split_name, context: 'foo' do |v|
  v.when :variant1, :variant2 do
    ...
  end
  v.default :variant1 do
    ...
  end
end
```

the behavior here is unclear for visitors assigned variant1, it's likely we should not allow the default to be specified in another `when`.

PR #38, opened 2017-10-24
add a warning notification message for calling vary with the same var…

effron:
/domain @samandmoore @aburgel @while1malloc0 @agirlnamedsophia 
/no-platform

Implementing this! https://github.com/Betterment/test_track_rails_client/issues/8

I figured we should start with a notify, and if this is live for a while without any issues, we can turn it into a raise.

I decided to implement this issue slightly more broadly: you should never run into a scenario where you need to define the same variant twice whether in `when` or `default`.

### DIFF
--- a/spec/models/test_track/vary_dsl_spec.rb
+++ b/spec/models/test_track/vary_dsl_spec.rb
@@ -169,11 +169,20 @@ RSpec.describe TestTrack::VaryDSL do
         expect(notifier).not_to have_received(:notify)
       end
     end
+
+    context "when adding a variant that was already configured" do
+      it "tells airbrake" do
+        subject.when :one, &noop
+        subject.when :one, &noop
+        expect(notifier).to have_received(:notify)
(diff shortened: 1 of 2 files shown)


spec/models/test_track/vary_dsl_spec.rb
```diff
@@ -169,11 +169,20 @@
...
+        expect(notifier).to have_received(:notify)
+          .with(/configures variant "one" more than once/)
+      end
+    end
   end
 
   context "#default" do
     it "accepts a block" do
-      subject.when :one do
+      subject.default :one do
```
effron: I also cleaned up this test case @samandmoore

samandmoore:
<<domain

samandmoore:
domainlgtm! thanks for adding this in. hopefully we don't have any cases where we do this.

samandmoore:
<<platformlgtm

effron:
nanda?

PR #39, opened 2017-11-27
Removing support for Mixpanel aliasing

asullivan210:
This PR decouples the TestTrack rails client from use of Mixpanel as an analytics tool

1. It removes the dependency on `ENV[MIXPANEL_TOKEN]`, which is currently called even if `TestTrack.analytics` is overridden to something other than the default MixpanelClient.
2. It removes the alias job and Mixpanel distinct id which are specific to Mixpanel's API, TestTrack will no longer be responsible for aliasing visitor id's (e.g. when a visitor logs in or signs up)

/domain @jmileham @aburgel @samandmoore 
/platform @jmileham @aburgel @samandmoore


smudge:
Looks like for JRuby tests we'll need to pin the `activerecord-jdbcsqlite3-adapter` gem version down to something like `~> 1.3.24`.

asullivan210:
adding specs and rebasing off the fixed build, will re-open

asullivan210:
bump

PR #40, opened 2017-11-27
Fix jruby builds

smudge:
/domain @asullivan210 @samandmoore @aburgel 
/no-platform

Pinning SQL adapter down to a version that should work.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,5 +1,6 @@
 appraise 'rails-4-2' do
   gem 'rails', '~> 4.2.0'
+  gem 'activerecord-jdbcsqlite3-adapter', '~> 1.3.24', platforms: :jruby
 end
 
 appraise 'rails-5-0' do
--- a/Appraisals
+++ b/Appraisals
@@ -6,12 +6,10 @@ appraise 'rails-5-0' do
   gem 'rails', '~> 5.0.0'
   gem 'rack'
   gem 'rails-controller-testing'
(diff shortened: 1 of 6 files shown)


smudge:
I lied -- one build still broken. Fixing that now.

aburgel:
<< domainlgtm

Appraisals
```diff
@@ -1,17 +1,16 @@
...
   gem 'rack'
   gem 'rails-controller-testing'
-  gem 'activerecord-jdbcsqlite3-adapter', github: 'jruby/activerecord-jdbc-adapter', branch: 'rails-5', platforms: :jruby
 end
 
 appraise 'rails-5-1' do
   gem 'rails', '~> 5.1.0'
   gem 'rack'
   gem 'rails-controller-testing'
-  gem 'activerecord-jdbcsqlite3-adapter', github: 'jruby/activerecord-jdbc-adapter', branch: 'rails-5', platforms: :jruby
```
samandmoore: @smudge why is it that we can remove this from here? we don't need this on rails 5+?

samandmoore: oh, i'm guessing that it doesn't work on rails 5+, correct?

samandmoore: Nope, I'm still confused.

smudge: I've left a generic line in the`Gemfile`, which the other appraisals will inherit:

```
gem 'activerecord-jdbcsqlite3-adapter', platforms: :jruby
```

samandmoore:
Ah. Yes yes. That’ll do it.

PR #41, opened 2017-11-27
Update to rubocop 0.51

aburgel:
/domain @jmileham @samandmoore 
/no-platform

This updates our rubocop version to 0.50 (can't go all the way to 0.51 because we still need to support ruby 2.0). I ran autocorrect to fix a bunch of the warnings.

In making this change, I updated a few of our other vendored dependencies.
1) I removed the private fork of `public_suffix` since we are now able to use the latest mainline version. 
2) I removed `ruby_spec_helpers`, as far as I could tell we were using it for the custom cops. Those are in the mainline 0.50 now.
3) Updated fakeable_her and her to the latest as well.


aburgel:
bump. changed the min version of ruby to 2.1.0 to ensure we have `YAML.safe_load`. I was also able to get on the latest version of rubocop with that change.

With @asullivan210's analytics PR landing soon, we'll need to do a major version bump anyways.

jmileham:
I thought Nathan’s note said there was a 2.0.x that had a functional safe_load? I was just figuring we could rock the tip of the 2.0 series. But if we’re about to make a breaking change anyway, it does seem not much is still supporting 2.0 so that’s fine too.

jmileham:
<< domain lgtm!

aburgel:
Turns out the latest release of 2.0 is 2.0.0-p648, there's no patch versions. 2.1 is the earliest version that includes `YAML.safe_load`, unless you upgrade the psych gem separately. Since we're making a breaking change shortly anyways, so I'm gonna merge.

smudge:
Latest release of 2.0.X is [2.0.14](http://blog.rubygems.org/2013/11/12/2.0.14-released.html). Dug around and it looks like `safe_load` was fixed in [this commit](https://github.com/ruby/psych/commit/95097bdacc48b7c6d6eeb875fed717d910be5603) which made it into 2.0.3.

aburgel:
@smudge that's rubygems, not the ruby runtime. TBH, I'm a little confused on how to think about ruby version support when the standard library can be overridden by a new version of rubygems. What are we committing to when we say we support 2.0?

smudge:
Right -- totally my own confusion. So, ruby semantic versioning didn't happen until 2.1.0. And that `safe_load` stuff went into the _psych_ repo, which I was assuming lined up with Ruby, but really they're just manually pulling changes into `ext/psych` in mainline Ruby. So it really does look like 2.1.0 is the first ruby version we can support with `safe_load`. Sorry, @aburgel, I will learn to read better 😅 .

PR #42, opened 2018-01-02
Arielle/remove mixpanel cookie setting

asullivan210:
/domain @aburgel 
/platform @aburgel @jmileham @samandmoore 

Today, consumers using an a non-Mixpanel analytics client are forced to define `ENV[MIXPANEL_TOKEN]` as TestTrack manages the mixpanel id regardless of whether the MixpanelClient is in use.

As of this PR, the TestTrack Rails Client will no longer provide functionality to alias TestTrack visitor identifiers with those of the analytics client, this functionality can still be implemented using a new `sign_up!` call back but the AliasJob and managing of `mixpanel_distinct_id` will no longer come with the TestTrack Rails Client

This is a breaking change and will involve releasing a new version, docs have been updated


spec/models/test_track/analytics/safe_wrapper_spec.rb
```diff
@@ -5,34 +5,18 @@
 
   subject { TestTrack::Analytics::SafeWrapper.new(underlying) }
 
-  describe '#alias' do
```
aburgel: Change this to a spec for `#sign_up!` along with the noop behavior if its not implemented?

aburgel:
<< domain

spec/rails_helper.rb
```diff
@@ -65,5 +65,4 @@
   config.include RailsCompatibilitySpecHelper, type: :controller
 end
 
-ENV['MIXPANEL_TOKEN'] = 'fakemixpaneltoken'
```
aburgel: I'm surprised we can remove this, since the mixpanel analytics code blows up without it.

spec/models/test_track/session_spec.rb
```diff
@@ -369,7 +366,7 @@ def current_clown; end
 
               expect(Thread).to have_received(:new)
               expect(TestTrack::UnsyncedAssignmentsNotifier).to have_received(:new) do |args|
-                expect(args[:mixpanel_distinct_id]).to eq("fake_distinct_id")
+                expect(args.keys).to contain_exactly :visitor_id, :assignments
```
aburgel: I think we can remove this line, we don't need to overly specify what args it expects.

spec/models/test_track/session_spec.rb
```diff
@@ -158,6 +148,50 @@ def current_clown; end
             )
           end
         end
+
+        context 'when analytics client is not default' do
```
aburgel: Ahh, I see the tests are here. I think we can move some of these tests to SafeWrapper since they're mostly testing its behavior.

aburgel:
Looking great! TAFN!

spec/rails_helper.rb
asullivan210: we never actually call `MixpanelClient#track_assignment` in the specs, we always stub it out

asullivan210:
bump

aburgel:
domainlgtm!

app/models/test_track/session.rb
```diff
@@ -64,9 +62,11 @@ def sign_up!(*args) # rubocop:disable Metrics/MethodLength
     end
```
jmileham: If we're going to 3.0 should we rip out this deprecated syntax support? Any other deprecation warnings like this?

jmileham:
Should this be bumping the gem version too?

<< platform tafn. Great stuff!

lib/test_track_rails_client/version.rb
```diff
@@ -1,3 +1,3 @@
 module TestTrackRailsClient
-  VERSION = "2.0.0" # rubocop:disable Style/MutableConstant
+  VERSION = "3.0.0" # rubocop:disable Style/MutableConstant
```
jmileham: Oh it does. Duh, i was looking for a gemspec diff.

app/models/test_track/session.rb
aburgel: 👍 I think that's the only one.

jmileham: (hopefully that gets us back under Rubocop's ABC/MethodLength thresholds too)

asullivan210:
bump, removed support for deprecated arguments to log_in and sign_up 👍

app/models/test_track/session.rb
```diff
@@ -30,18 +30,12 @@ def state_hash
...
-      identifier_type = identity.test_track_identifier_type
-      identifier_value = identity.test_track_identifier_value
-    else
-      identifier_type = args[0]
-      identifier_value = args[1]
-      warn "#log_in! with two args is deprecated. Please provide a TestTrack::Identity"
-    end
+    identity = args[0]
+    identifier_type = identity.test_track_identifier_type
+    identifier_value = identity.test_track_identifier_value
```
jmileham: I think we no longer use opts, and we can just name the args individually instead of taing *args, right?

jmileham: oh we use opts. Are we ruby 2-only yet? We could use a named optional arg instead of opts too.

asullivan210: yup we're `>= 2.1.0`

asullivan210:
retest this please

asullivan210:
bump

asullivan210:
bump

app/models/test_track/session.rb
```diff
@@ -31,42 +30,27 @@ def state_hash
...
-    if args[0].is_a?(TestTrack::Identity)
-      identity = args[0]
-      identifier_type = identity.test_track_identifier_type
-      identifier_value = identity.test_track_identifier_value
-    else
-      identifier_type = args[0]
-      identifier_value = args[1]
-      warn "#log_in! with two args is deprecated. Please provide a TestTrack::Identity"
-    end
+  def log_in!(identity, opts = {})
```
jmileham: I think we can make this a `forget_current_vistor: nil` kwarg instead of opts

jmileham:
tiny opts tweake and platform lgtm

PR #44, opened 2018-04-13
Don't sync assignment events for gates

### LINKED ISSUE #43: Stop recording assignments for feature gates
Per this comment on test_track server: https://github.com/Betterment/test_track/pull/69#issuecomment-380995245

Since gates (splits having a name with a suffix of `_enabled`) are not used for analysis, we can save storage and increase the fluidity of release/rollback operations by not storing assignments at all for feature gates. If we plan not to store them, then we can also skip sending segment and testtrack assignment events.

A byproduct of not storing them is an increased amount of assignment event chattiness from clients that will never realize that a visitor has already been assigned a variant for a given split. This will be a fast event for TestTrack to serve because it will quickly realize that there's no work to do, but we might as well kill the event sending from the client and avoid confusion about duplicate assignment events.

jmileham:
Fixes #43 

Because feature gates aren't needed for A/B test analysis and [TestTrack server will now discard the assignment event](https://github.com/Betterment/test_track/pull/69
) in order to allow visitors to dynamically reassign upon gate weighting changes, we just deterministically assign locally each time instead and skip notification of the event.

This code change is pretty far from the notification code, but leans on the fact that all notification codepaths rely on visitor.unsynced_assignments to decide whether to send assignment notifications, which in turn calls the `unsynced?` method on assignment, which was previously hard-coded to return true - basically if the rails client made an assignment, it was considered worthy of syncing to the server before. Now we will consider feature gate assignments to be synced, so they will not trigger a notification.

This should make pages and jobs that only have feature gates (and not experiment assignments) less resource intensive than ever!

### DIFF
--- a/app/models/test_track/analytics/mixpanel_client.rb
+++ b/app/models/test_track/analytics/mixpanel_client.rb
@@ -1,7 +1,7 @@
 module TestTrack::Analytics
   class MixpanelClient
     def track_assignment(visitor_id, assignment)
-      mixpanel.track(visitor_id, 'SplitAssigned', split_properties(assignment).merge(TTVisitorID: visitor_id))
+      mixpanel.track(visitor_id, event_name(assignment), split_properties(assignment).merge(TTVisitorID: visitor_id))
     end
 
     private
@@ -9,6 +9,14 @@ module TestTrack::Analytics
     def mixpanel
       raise "ENV['MIXPANEL_TOKEN'] must be set" unless ENV['MIXPANEL_TOKEN']
       @mixpanel ||= Mixpanel::Tracker.new(ENV['MIXPANEL_TOKEN'])
+    end
+
+    def event_name(assignment)
+      if assignment.feature_gate?
+        'FeatureGateExperienced'
+      else
+        'SplitAssigned'
+      end
     end
 
     def split_properties(assignment)
--- a/app/models/test_track/analytics/mixpanel_client.rb
+++ b/app/models/test_track/analytics/mixpanel_client.rb
@@ -1,7 +1,7 @@
 module TestTrack::Analytics
   class MixpanelClient
-    def track_assignment(visitor_id, assignment)
-      mixpanel.track(visitor_id, event_name(assignment), split_properties(assignment).merge(TTVisitorID: visitor_id))
+    def track(analytics_event)
+      mixpanel.track(analytics_event.visitor_id, analytics_event.name, analytics_event.properties)
     end
 
     private
@@ -9,22 +9,6 @@ module TestTrack::Analytics
     def mixpanel
       raise "ENV['MIXPANEL_TOKEN'] must be set" unless ENV['MIXPANEL_TOKEN']
       @mixpanel ||= Mixpanel::Tracker.new(ENV['MIXPANEL_TOKEN'])
-    end
-
-    def event_name(assignment)
-      if assignment.feature_gate?
-        'FeatureGateExperienced'
-      else
-        'SplitAssigned'
-      end
-    end
-
-    def split_properties(assignment)
-      {
-        SplitName: assignment.split_name,
-        SplitVariant: assignment.variant,
-        SplitContext: assignment.context
-      }
     end
   end
 end
--- a/app/models/test_track/assignment.rb
+++ b/app/models/test_track/assignment.rb
@@ -16,7 +16,11 @@ class TestTrack::Assignment
   end
 
   def unsynced?
-    true
+    !feature_gate?
+  end
+
+  def feature_gate?
+    split_name.end_with?('_enabled')
   end
 
   private
--- a/app/models/test_track/assignment.rb
+++ b/app/models/test_track/assignment.rb
@@ -23,6 +23,14 @@ class TestTrack::Assignment
     split_name.end_with?('_enabled')
   end
 
+  def analytics_event
+    @analytics_event ||= AnalyticsEvent.new(self)
+  end
+
+  def visitor_id
+    visitor.id
+  end
+
   private
 
   def _variant
--- a/app/models/test_track/assignment.rb
+++ b/app/models/test_track/assignment.rb
@@ -16,7 +16,7 @@ class TestTrack::Assignment
   end
 
   def unsynced?
-    !feature_gate?
+    true
   end
 
   def feature_gate?
--- a/spec/models/test_track/analytics/safe_wrapper_spec.rb
+++ b/spec/models/test_track/analytics/safe_wrapper_spec.rb
@@ -3,20 +3,22 @@ require 'rails_helper'
 RSpec.describe TestTrack::Analytics::SafeWrapper do
   let(:underlying) { double("Client", track: true, sign_up!: true) }
 
+  let(:event) { instance_double(TestTrack::AnalyticsEvent) }
+
   subject { TestTrack::Analytics::SafeWrapper.new(underlying) }
 
   describe '#track' do
     it 'calls underlying' do
-      expect(subject.track("my event")).to eq true
-      expect(underlying).to have_received(:track).with("my event")
+      expect(subject.track(event)).to eq true
+      expect(underlying).to have_received(:track).with(event)
     end
 
     context 'underlying raises' do
       it 'returns false' do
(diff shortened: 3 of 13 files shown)


app/models/test_track/assignment.rb
```diff
@@ -16,7 +16,11 @@ def variant
   end
 
   def unsynced?
-    true
+    !feature_gate?
+  end
+
+  def feature_gate?
+    split_name.end_with?('_enabled')
```
jmileham: The wire protocol TestTrack uses for split registries is too concise to add an explicit field describing gate-iness without changing it completely. Currently it's like so:

```JSON
{
  "split_name": { "variant1": 50, "variant2": 50 }
}
```

So I'm planning on making `_enabled` a hard requirement of feature gate naming as we advance from here so that the client can reliably infer gate-iness this way.

jmileham:
Note that this affects both metrics collection assignment notifications and TT server notifications. @asullivan210 I had been misremembering the way the rails client worked. I thought it sent an experience event too, but realized now that TT doesn't consider itself responsible for experience events, only assignment events.

So this change will make feature gate events go dark unless you instrumented your app code per-feature. At the same time, continuing to send assignment events for feature gates as we were seems like a lie, because they'd become too chatty. Maybe we should think about adding a split experienced event *instead*? Or maybe we can just use generalized app analytics to track traffic on the page we know to be gated if we want to know how much traffic its getting?

jmileham:
/domain joejansen given this comment: https://github.com/Betterment/test_track/pull/69#comment-181493116 you’ll definitely have a horse in this race then.

This PR doesn’t block the TT server PR so we can mess with this til it does satisfying things.

app/models/test_track/assignment.rb
```diff
@@ -16,7 +16,11 @@ def variant
   end
 
   def unsynced?
```
samandmoore: i think we should rename this to `should_sync?`. since some `Assignment`s are no longer meant to be synced, it seems to me that this name is misleading as is. i.e. feature gates are intentionally "unsynced" so changing the naming here will provide a clearer indication of that.

samandmoore:
<<domainlgtm

i reasoned through this for a while, and this seems to be the least invasive and simplest way to change the behavior. 

just change that method name 👍

jmileham:
Cool so the last bit is how we want to ensure devs get a sense for how many people are experiencing which fork of a gate. Sam and I were thinking maybe there’s another event we want to send to the event bus indicating that the client selected variant X of feature gate Y. @joejansen thoughts?

joejansen:
i think there is value in having some analytics event sent to segment or whatever analytics client the test track clients configure. when we ship something that relies on a feature gate, an analytics event can help us confirm that things are working as we intend and in the proportion or absolute volume we'd expect. in many cases we use feature gates in model code that will be triggered by a delayed job or a system transaction where there is no other analytics event that can serve as a good proxy. further, even when there is a proxy event, we may be relying on some filtering logic that only exposes a portion of users to the gate based on state the proxy event is unaware of. 

i understand the event name `SplitAssigned` no longer fits as well given that we are no longer recording an assignment in the server application. i'd be in favor of implementing a new analytics event like `SplitExperienced` or something even more specific to this use case like `FeatureFlagExperienced` or `FeatureFlagExercised`.

joejansen:
<<domain tafn

jmileham:
bump. Total change of approach here.

Now for feature gates we skip TT assignment events but continue to send an analytics event. The built-in mixpanel client now sends a `FeatureGateExperienced` event instead of a `SplitAssigned` event. Consuming apps will have to modify their custom analytics clients to differentiate event types like this, but if they fail to, they'll just get noisey assignments for longer. Not the end of the world.

 @samandmoore you'll probably want to give this another look.

app/models/test_track/analytics/mixpanel_client.rb
```diff
@@ -11,6 +11,14 @@ def mixpanel
       @mixpanel ||= Mixpanel::Tracker.new(ENV['MIXPANEL_TOKEN'])
     end
 
+    def event_name(assignment)
+      if assignment.feature_gate?
+        'FeatureGateExperienced'
+      else
+        'SplitAssigned'
+      end
+    end
```
jmileham: I could stuff this function into assignment as a method, e.g. `analytics_event_name`. It'd be your option whether you choose use it in your custom analytics plugin or use something else but you wouldn't have to duplicate the biz logic then.

jmileham:
bump - I revoked your LGs. Sorry. But I changed too much not to get another look.

Ok, so this just became a significantly breaking change to the custom analytics client API contract (like releasing client v4.0-level breaking change), but I think that's OK because you'd want to get on board with the event name changes for feature gates anyway, and not having a forcing function to reassess when upgrading the TT client would probably result in confusion.

After talking with @samandmoore offline I changed the `AnalyticsClient#track_assignment(visitor_id, assignment)` contract to `AnalyticsClient#track(analytics_event)`. `TestTrack::AnalyticsEvent` provides the business logic to alternate between feature gate/experiment event names and provides a ready-to-use property payload so that the analytics client merely needs to figure out how to instantiate the tracking lib and call it with the right args instead of do a bunch of data munging and hash construction per plugin.

I also updated the readme to match and dropped upgrading to 4.0 instructions.

I don't actually intend to release this as 4.0 until we also make multi-identity sessions a thing in order to allow lookups of variants of other people from model code.

jmileham:
nanda?

spec/models/test_track/analytics_event_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      )
+    end
+
+    subject { described_class.new(assignment) }
+
+    it "has a name of SplitAssigned" do
+      expect(subject.name).to eq "SplitAssigned"
+    end
+  end
+end
```
samandmoore: Do we care about documenting the public interface of this class? For example, that it exposes assignment and visitor_id as attrs. 

Since we don’t have api docs, the only real proof that we won’t rip out some public readers is a test case. 

It’s sorta documented through the other test that uses this object as a double, but that feels pretty indirect.

spec/models/test_track/analytics/safe_wrapper_spec.rb
```diff
@@ -1,22 +1,22 @@
...
+  let(:underlying) { double("Client", track: true, sign_up!: true) }
 
   subject { TestTrack::Analytics::SafeWrapper.new(underlying) }
 
-  describe '#track_assignment' do
+  describe '#track' do
     it 'calls underlying' do
-      expect(subject.track_assignment(123, foo: "bar")).to eq true
-      expect(underlying).to have_received(:track_assignment).with(123, foo: "bar")
+      expect(subject.track("my event")).to eq true
```
samandmoore: We don’t care that this is no longer a true representation of the contract do we? I think it’s fine.

spec/models/test_track/assignment_spec.rb
```diff
@@ -48,4 +48,18 @@
...
+    end
+
+    context "for a feature gate" do
+      subject { described_class.new(visitor: visitor, split_name: :feature_enabled) }
+
+      it "also returns true" do
+        expect(subject.unsynced?).to eq true
+      end
+    end
+  end
```
samandmoore: Keeping this test around for clarity?

samandmoore:
<<domainlgtm

spec/models/test_track/analytics/safe_wrapper_spec.rb
jmileham: do you mean you'd prefer I use a mock that looks more like an event?

spec/models/test_track/assignment_spec.rb
jmileham: yeah, now that we've decided that `unsynced?` means something else, it felt good to leave this test to lock it down. I also discovered that the fake assignment does leverage this to return false, so everything was getting gnarly.

spec/models/test_track/analytics_event_spec.rb
jmileham: cool, i reorged the tests around typical `describe "#method_name"` style and hit all the public interface.

jmileham:
bump @joejansen this cool w/you?

PR #46, opened 2018-04-26
Manage multiple identities within a web session

jmileham:
/domain @samandmoore @smudge 
/cc @noveng05

I haven't tried this out in anger yet, but the tests pass and I wanted to get some feedback.

This is not a perfect factoring. Ideally we can decouple sessions from controllers more (so we can make offline sessions also fully stateful and employ RequestStore) and not special case the currently signed in identity so hard.

### DIFF
--- a/app/models/test_track/identity_session_locator.rb
+++ b/app/models/test_track/identity_session_locator.rb
@@ -1,4 +1,4 @@
-class TestTrack::IdentitySessionDiscriminator
+class TestTrack::IdentitySessionLocator
   attr_reader :identity
 
   def initialize(identity)
@@ -8,8 +8,8 @@ class TestTrack::IdentitySessionDiscriminator
   def with_visitor
     raise ArgumentError, "must provide block to `with_visitor`" unless block_given?
 
-    if matching_identity?
-      yield session.visitor_dsl
+    if web_context?
+      yield session.visitor_dsl_for(identity)
     else
       TestTrack::OfflineSession.with_visitor_for(identity.test_track_identifier_type, identity.test_track_identifier_value) do |v|
         yield v
@@ -28,10 +28,6 @@ class TestTrack::IdentitySessionDiscriminator
   end
 
   private
-
-  def matching_identity?
-    session.present? && session.has_matching_identity?(identity)
-  end
 
   def web_context?
     session.present?
--- a/app/models/test_track/session.rb
+++ b/app/models/test_track/session.rb
@@ -16,6 +16,27 @@ class TestTrack::Session
     notify_unsynced_assignments! if sync_assignments?
   end
 
+  def visitor_dsl_for(identity)
+    if has_matching_identity?(identity)
+      visitor_dsl
+    else
+      TestTrack::VisitorDSL.new(visitors_by_identity[identity])
+    end
+  end
+
+  def visitors_by_identity
+    @visitors_by_identity ||= Hash.new do |visitors_by_identity, identity|
+      remote_visitor = TestTrack::Remote::Visitor.from_identifier(
+        identity.test_track_identifier_type,
+        identity.test_track_identifier_value
+      )
+      visitors_by_identity[identity] = TestTrack::Visitor.new(
+        id: remote_visitor.id,
+        assignments: remote_visitor.assignments
+      )
+    end
+  end
+
   def visitor_dsl
     @visitor_dsl ||= TestTrack::VisitorDSL.new(visitor)
   end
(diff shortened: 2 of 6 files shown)


app/models/test_track/identity_session_locator.rb
```diff
@@ -8,8 +8,8 @@ def initialize(identity)
   def with_visitor
     raise ArgumentError, "must provide block to `with_visitor`" unless block_given?
 
-    if matching_identity?
-      yield session.visitor_dsl
+    if web_context?
+      yield session.visitor_dsl_for(identity)
```
jmileham: Eventually I think we'll require that a session be present whenever TestTrack splits are exercised and this whole web context thing will go away, and it'll naturally be wired into job lifecycles, etc. But for now, the key change here is that if you've got a TT session (you're in a "web context"), we're going to expect the online session to cough up a TT visitor for us - even if it's not the same identity that is logged into the web.

app/models/test_track/session.rb
```diff
@@ -16,6 +16,27 @@ def manage
     notify_unsynced_assignments! if sync_assignments?
   end
 
+  def visitor_dsl_for(identity)
+    if has_matching_identity?(identity)
+      visitor_dsl
```
jmileham: This fork is still a little ugly, because I didn't want to refactor too hard in this pass. Ultimately I'd like the session's primary identity to be just like any other identity we execute splits on.

app/models/test_track/session.rb
```diff
@@ -16,6 +16,27 @@ def manage
...
+    @visitors_by_identity ||= Hash.new do |visitors_by_identity, identity|
+      remote_visitor = TestTrack::Remote::Visitor.from_identifier(
+        identity.test_track_identifier_type,
+        identity.test_track_identifier_value
+      )
+      visitors_by_identity[identity] = TestTrack::Visitor.new(
+        id: remote_visitor.id,
+        assignments: remote_visitor.assignments
+      )
+    end
```
jmileham: So I think this is clever but in a not awful way - using a custom Hash initialization to auto-add to the collection whenever somebody requests a visitor for an identity that we've never seen before.

Maybe this should be private so nobody does anything naughty to the hash's state from outside the model.

spec/models/test_track/identity_session_locator_spec.rb
```diff
@@ -16,36 +16,12 @@
...
-
-        it "creates an offline session and yields its visitor" do
-          subject.with_visitor do |visitor|
-            expect(visitor).to eq visitor_dsl
-          end
-
-          expect(TestTrack::OfflineSession).to have_received(:with_visitor_for).with("clown_id", 1234)
+      it "yields the session's visitor dsl" do
+        subject.with_visitor do |visitor|
+          expect(visitor).to eq visitor_dsl
```
jmileham: This test simplification brought to you by supporting multiple identities.

spec/models/test_track/session_spec.rb
```diff
@@ -422,6 +422,60 @@ def current_clown; end
...
+      expect(subject.visitors_by_identity[identity]).to be_a TestTrack::Visitor
+
+      expect(TestTrack::Remote::Visitor).to have_received(:from_identifier).with("foo_user_id", "123")
+    end
+
+    it "only fetches a remote visitor once for the same identity" do
+      subject.visitors_by_identity[identity]
+      subject.visitors_by_identity[identity]
+
+      expect(TestTrack::Remote::Visitor).to have_received(:from_identifier).with("foo_user_id", "123").exactly(:once)
```
jmileham: It's kinda nice to test this separately, but I don't think it'd be very gnarly to test this through `visitor_dsl_for`. Of course the real pro move would be to extract to another class. But I'd like to do that as I get rid of the IdentityCollection class which is really confusing.

app/models/test_track/session.rb
samandmoore: my vote is to extract this Hash into a class that has an easier-to-understand interface than the self-populating hash thing we're doing here. i'm fine with waiting to do that until we get rid of the `SessionIdentityCollection` so that you can avoid any confusion about what that class is vs. this new one i'm proposing.

samandmoore:
i've been through this a few times now. i like where this is going. i'm excited to see the Session class get less complicated :)

<<domainlgtm!

jmileham:
ok, i'm going to merge this. being a gem, it's ok if it's not thoroughly tested the second it hits master. we can do some testing from there and that'll allow me to build on top of it. If this does what we expect in integration testing, we can deploy it to rubygems as a 4.0alpha release or something and start depending on it for real.

samandmoore:
That sounds like a good plan. We can also depend on it from source via
GitHub if we want to start testing sooner 🙂 alpha release is valuable if
other teams wanna test it through the usual channels.

PR #47, opened 2018-04-28
Bump version to 4.0.0-alpha1

jmileham:
Functionally improves on 3.0.1 (skips sending assignments to TT for feature gates, upgrades analytics client contract) but we're not done yet with 4.0.0 features (fully managed offline sessions)

/domain @samandmoore 
/no-platform


samandmoore:
<<domainlgtm

Just double check that that’s the right format for an alpha on ruby gems and sgtm 💯

jmileham:
I’ll double check

jmileham:
rubygems docs indicate a dot instead of a hyphen at least conventionally. I guess I'll go with that.

PR #48, opened 2018-04-29
Conform to our latest event property naming

jmileham:
/no-platform

Use snake case params by default now so our custom clients don't need to reimplement.


samandmoore:
<<domainlgtm cool. that always bugged me.

PR #49, opened 2018-04-30
snake-ize event names too

jmileham:

### DIFF
--- a/app/models/test_track/analytics_event.rb
+++ b/app/models/test_track/analytics_event.rb
@@ -10,9 +10,9 @@ module TestTrack
 
     def name
       if assignment.feature_gate?
-        'FeatureGateExperienced'
+        'feature_gate_experienced'
       else
-        'SplitAssigned'
+        'split_assigned'
       end
     end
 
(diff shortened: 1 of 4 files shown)


app/models/test_track/analytics_event.rb
```diff
@@ -10,9 +10,9 @@ def initialize(assignment)
 
     def name
       if assignment.feature_gate?
-        'FeatureGateExperienced'
+        'feature_gate_experienced'
       else
-        'SplitAssigned'
+        'split_assigned'
```
noveng05: do these mixpanel [specs](https://github.com/Betterment/test_track_rails_client/blob/master/spec/models/test_track/analytics/mixpanel_client_spec.rb#L12) need some updating too?

Also, just curious whether this "breaks" the continuity of some funnels being tracked in mixpanel?  Do folks need to update what they're tracking to look at either `SplitAssigned` or `split_assigned` when this lands in the apps?

noveng05:
<< domaintafn
just a couple of questions to better my understanding

jmileham:
Yeah, so the context is I failed to snake-ize the event names when i snake-ized the property names. `split_experienced` is the event name we already use these days, just making TestTrack generate that name by default rather than have to override it elsewhere. I'll release alpha3 and then upgrade to use this once it lands.

noveng05:
got it. thanks for the context. 
domainlgtm

app/models/test_track/analytics_event.rb
jmileham: I can update these specs too though they're not actually hurting anything, just a little less realistic.

jmileham:
This passes locally gonna merge optimistically.

PR #50, opened 2018-04-30
Fix module resolution bug for TestTrack::AnalyticsEvent

jmileham:
All of the alphas. This simple memoized getter never worked, and I didn't test it. Womp. Cutting alpha4 off the back of this too.

/no-platform


jmileham:
yolomerge again. tests pass.

samandmoore:
<< domain lgtm

PR #52, opened 2018-05-01
Clean up assignment interface

jmileham:
/domain @noveng05 @samandmoore 
/no-platform

urrrrg I keep finding more breakage. This removes the Assignment#analytics_event method so we can duck type it like I should've done last version.

I think I'm close to the bottom of the rabbit hole. This is the only thing currently generating errbits in stage.


noveng05:
<< domainlgtm

PR #53, opened 2018-05-01
AnalyticsAssignments need to say whether they're feature gates

jmileham:
/no-platform

It keeps getting better and hackier. Maybe this is the bottom. I wish the testing cycle were better on this jazz. Now we're deciding which analytics event name to send at last. Getting close.


samandmoore:
<<domainlgtm

PR #54, opened 2018-05-04
Use identity if authenticated

jmileham:
/domain @aburgel @samandmoore 
/platform @aburgel @samandmoore 

This is a pretty significant change to web session management.

Now, instead of trusting the cookie/header-provided visitor_id, if there's an authenticated identity, we'll ignore the cookie/header visitor_id and use the authenticated identity instead. This will provide a strong guarantee that the visitor is who they should be. Notes below.


README.md
```diff
@@ -51,15 +55,27 @@ Set up ENV vars in every environment:
...
+  self.test_track_identity = :current_user
 end
 ```
 
+### Prepare your identity models
+
+You'll need to configure your `User` model as a [TestTrack Identity](#varying-app-behavior-from-within-a-model)
+
+If your app doesn't support authentication, just define a method that
+always returns `nil`, e.g. `def current_user; end`.
```
jmileham: This just became non-optional because it's the primary mechanism for visitor lookup in an authenticated state, so I added it to the installation docs.

README.md
```diff
@@ -239,51 +255,9 @@ if test_track_visitor.ab :dark_deployed_feature, context: 'signup'
...
 
-### Varying app behavior in an offline context
-
-The `OfflineSession` class can be used to load a test track visitor when there is no access to browser cookies. It is perfect for use in a process being run from either a job queue or a scheduler. The visitor object that is yielded to the block is the same as the visitor in a controller context; it has both the `vary` and `ab` methods.
-
-An `OfflineSession` can be established in one of two ways:
-
-1. with an `identifier_type`:
-```ruby
-OfflineSession.with_visitor_for(:myapp_user_id, 1234) do |test_track_visitor|
```
jmileham: I decided that we should privatize this OfflineSession stuff and lean into the Identity concept. It is much easier to work with and allows us to manage the session better when we get to that part. So I ripped the docs out of the README.

README.md
```diff
@@ -239,51 +255,9 @@ if test_track_visitor.ab :dark_deployed_feature, context: 'signup'
...
-    end
-  end
-end
-```
-
 ### Varying app behavior from within a model
 
-The `TestTrack::Identity` concern can be included in a model and it will add two methods to the model: `test_track_vary` and `test_track_ab`. Behind the scenes, these methods check to see if they are being used within a web context of a controller that includes `TestTrack::Controller` or not. If called in a web context they will use the `test_track_visitor` that the controller has and participate in the existing session, if not, they will standup an `OfflineSession`.
-
-Because these methods may need to stand up an `OfflineSession` the consuming model needs to provide both the identifier type and which column should be used as the identifier value via the `test_track_identifier` method so that the `OfflineSession` can grab the correct visitor.
```
jmileham: This is an implementation detail that I think will just cofuse users, and may change as we refactor offline sessions for visitor reuse.

README.md
```diff
@@ -317,7 +289,7 @@ The `test_track_visitor.sign_up!` method tells TestTrack when a new identifier h
 test_track_visitor.sign_up!(:myapp_user_id, 2345)
 ```
 
-## Testing splits
+## Testing your split-dependent application code with RSpec
```
jmileham: Since we're doing split testing, I thought this could benefit from a little clarification.

README.md
```diff
@@ -384,6 +356,10 @@ easier and more conventional, though, and takes care of differentiating
 between expiriment assignments and feature gate experiences, which are
 no longer recorded server-side.
 
+You also must add `self.test_track_identity = :current_user` (or
+whatever your controller uses as a sign-in identity) to your
+TestTrack-enabled controllers.
```
jmileham: Upgrade notes for the new requirement.

app/controllers/concerns/test_track/controller.rb
```diff
@@ -2,6 +2,8 @@ module TestTrack::Controller
   extend ActiveSupport::Concern
 
   included do
+    class_attribute :test_track_identity
```
jmileham: I decided to use a class_attribute for this because it does inheritance correctly out of the box regardless of the crazy controller hierarchies people come up with.

app/models/test_track/ab_configuration.rb
```diff
@@ -19,7 +19,9 @@ def variants
   private
 
   def build_variant_hash
-    notify_because_ab("configures split with more than 2 variants") if split_variants && split_variants.size > 2
+    if split_variants && split_variants.size > 2 # rubocop:disable Style/SafeNavigation
```
jmileham: rubocop is wrong about this cop here. what else is new, though.

app/models/test_track/lazy_visitor_by_identity.rb
```diff
@@ -0,0 +1,38 @@
+class TestTrack::LazyVisitorByIdentity
```
jmileham: This model is a little interesting. `TestTrack::Visitor` objects will lazily fetch on demand, but always have an `id` attribute. When fetching a visitor by identity, that's no longer true - you actually fetch the visitor ID from TT so even retrieving `id` would require an API call.

So I added the concept of `id_loaded?` in addition to Visitor's existing concept of `loaded?`. As I said, regular visitors can always say `id_loaded => true`. But this lazy visitor will return `false` until it's actually fetched.

There are some housekeeping operations that aren't worth fetching `id` just to do them, but we'll gladly do them if the `id` is already loaded, so that's the rationale here.

For this model, `loaded?` and `id_loaded?` are the same, but for a regular `Visitor`, `loaded?` will return true only after the remote visitor has been fetched from the server to get the assignment bag.

app/models/test_track/lazy_visitor_by_identity.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def id_loaded?
+    loaded?
+  end
+
+  private
+
+  def method_missing(method, *args, &block) # rubocop:disable Style/MethodMissing
+    __visitor__.send(method, *args, &block)
+  end
```
jmileham: Rubocop is wrong here too. If you want to delegate all the things, you don't want to do the method-whitelist-else-super dance they suggest. SimpleDelegator wouldn't be a thing if that advice were universal.

app/models/test_track/session.rb
```diff
@@ -11,34 +11,25 @@ def initialize(controller)
   def manage
     yield
   ensure
-    manage_cookies!
-    manage_response_headers!
+    if visitor.id_loaded?
+      manage_cookies!
+      manage_response_headers!
+    end
```
jmileham: These are the operations that aren't worth doing if we don't actually use the visitor. Only possible bad outcome I can imagine from this would be that the cookie never ever gets an updated expiry timestamp if we never use TestTrack in an authenticated controller *and* the user never experiences an unauthenticated page for a whole year. Not likely.

app/models/test_track/session.rb
```diff
@@ -11,34 +11,25 @@ def initialize(controller)
...
     notify_unsynced_assignments! if sync_assignments?
```
jmileham: `sync_assignments?` already depends on `loaded?`

app/models/test_track/session.rb
```diff
@@ -11,34 +11,25 @@ def initialize(controller)
...
   end
 
   def visitor_dsl_for(identity)
-    if has_matching_identity?(identity)
-      visitor_dsl
-    else
-      TestTrack::VisitorDSL.new(visitors_by_identity[identity])
-    end
+    TestTrack::VisitorDSL.new(visitors_by_identity[identity])
```
jmileham: Now we're not special casing the identity-that-matches-the-controller-user - all visitor's being located by identity hit the same endpoint. This is the crux of the change - never trust a relationship between a visitor_id from the controller and an authenticated identity.

app/models/test_track/session.rb
```diff
@@ -11,34 +11,25 @@ def initialize(controller)
...
     @visitors_by_identity ||= Hash.new do |visitors_by_identity, identity|
-      remote_visitor = TestTrack::Remote::Visitor.from_identifier(
-        identity.test_track_identifier_type,
-        identity.test_track_identifier_value
-      )
-      visitors_by_identity[identity] = TestTrack::Visitor.new(
-        id: remote_visitor.id,
-        assignments: remote_visitor.assignments
-      )
+      visitors_by_identity[identity] = TestTrack::LazyVisitorByIdentity.new(identity)
```
jmileham: This tidies up by delegating to the lazy visitor.

app/models/test_track/session.rb
```diff
@@ -11,34 +11,25 @@ def initialize(controller)
...
     end
   end
 
   def visitor_dsl
-    @visitor_dsl ||= TestTrack::VisitorDSL.new(visitor)
+    TestTrack::VisitorDSL.new(visitor)
```
jmileham: There's no need to be caching this object - it's stateless.

app/models/test_track/session.rb
```diff
@@ -52,41 +43,49 @@ def state_hash
...
 
   def log_in!(identity, forget_current_visitor: nil)
-    identifier_type = identity.test_track_identifier_type
-    identifier_value = identity.test_track_identifier_value
-
-    @visitor = TestTrack::Visitor.new if forget_current_visitor
-    visitor.link_identifier!(identifier_type, identifier_value)
+    @unauthenticated_visitor = TestTrack::Visitor.new if forget_current_visitor
+    unauthenticated_visitor.link_identity!(identity)
+    visitors_by_identity[identity] = unauthenticated_visitor
```
jmileham: I changed the contract of `visitor.link_identifier!` to `visitor.link_identity!`. Identities work nice. Let's use them.

samandmoore:
Where’d we end up on this?

README.md
samandmoore: Where does this method need to be defined?

app/models/test_track/lazy_visitor_by_identity.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def respond_to_missing?(method, include_private = false)
+    super || visitor.respond_to?(method, include_private)
```
samandmoore: Is `visitor` correct here or is this supposed to be __ or @?

app/models/test_track/visitor.rb
```diff
@@ -93,12 +98,27 @@ def id_overridden_by_existing_visitor?
 
   private
 
+  def identifier_opts(identity)
+    {
+      identifier_type: identity.test_track_identifier_type,
+      visitor_id: id, value: identity.test_track_identifier_value.to_s
```
samandmoore: I think you’re missing a newline here

app/models/test_track/session.rb
```diff
@@ -52,41 +43,49 @@ def state_hash
...
+    end
+  end
+
+  def current_identity
+    raise <<~ERROR unless controller.class.test_track_identity&.is_a?(Symbol)
+      Your controller (or controller base class) must set test_track_identity for
+      TestTrack to work properly. e.g.:
+
+        self.test_track_identity = :current_user
+    ERROR
```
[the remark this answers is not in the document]
samandmoore: If we want to avoid the requirement of specifying an empty method, we could document a special symbol :none that we handle here as nil automatically.

It’s a small thing that would provide a nicer experience for anyone that does happen to use this in a system that doesn’t have auth

README.md
samandmoore: We should mention that this goes in the controller.

app/models/test_track/session.rb
samandmoore: I still think it would be nice to have an object instead of relying on the Hash default proc thing here.

samandmoore:
<<domain tafn

app/models/test_track/session.rb
jmileham: Ok, I’ll try that. The lazy visitor has enough responsibilities though and I thought this got skinny enough to be idiomatic as a hash default, but it’s still pretty heady.

app/models/test_track/session.rb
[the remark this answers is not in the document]
jmileham: Yeah, I spose nobody’s got a signed-in-entity getter called `#none`

app/models/test_track/lazy_visitor_by_identity.rb
jmileham: Whoops. __ it is.

app/models/test_track/visitor.rb
jmileham: Yurp

jmileham:
bump! I ended up going all the way to the bottom of that hash refactor and it slimmed down session a lot, so that's good.

samandmoore:
domainlgtm! The new changes are awesome. The session logic is much clearer now. 👏🏻

I’m fine with ditching 2.1. Modern deploy tooling makes it easy to use newer rubies.

jmileham:
BTW re this comment:

> I’d love to find a way to tie into the apps authentication mechanism to avoid having to load the visitor from TT on every web request back to the app server.

As we talked about offline, I don't think it'd be desirable to avoid loading the visitor from TT on every web request - that would add cache invalidation bugs if somebody got a new override assignment - but it would be desirable to allow fallback to a cached visitor + assignment registry stored in the rails session (so it has the same authentication guarantees as the web session) if any given identifier_visitor fetch times out.

spec/models/test_track/session_spec.rb
```diff
@@ -306,6 +306,21 @@ def current_clown; end
           expect(unsynced_assignments_notifier).to have_received(:notify)
         end
 
+        it "notifies unsynced assignments for identities" do
+          subject.manage do
+            subject.visitor_dsl_for(identity).ab('bar', true_variant: 'baz', context: :spec)
+          end
+
+          expect(Thread).to have_received(:new)
+          expect(TestTrack::UnsyncedAssignmentsNotifier).to have_received(:new) do |args|
```
samandmoore: this assertion might be flaky if the thread doesn't run immediately. in the old test case we were awkwardly grabbing a handle on the thread and joining it. maybe this spec should just assert that `ThreadedVisitorNotifier` is called with the right args and the tests against `ThreadedVisitorNotifier` can do the thread#join stuff.

samandmoore: and you can remove the `Thread` assertion from here as well.

jmileham: I wanted to preserve this as an integration test. This is in the same block as above and Thread.new is inlined - couldn’t spy on it if there weren’t funny business going on. This shouldn’t be flakey.

samandmoore: oh i see! jeeze, rspec is the worst for test-setup-clarity sometimes. carry on.

reustmd:
Did we gain more insight into why we were seeing so many misleading headers/cookies?

PR #57, opened 2018-05-25
Fix analytics tracking errors by providing visitor_id for remote assignments

aburgel:
/no-platform
/cc @wcirillo @jmileham 

We've been seeing a lot of error logging in our analytics code related to assignments with a missing visitor id. I think its caused by a difference between the remote and local assignment classes.

Both have a `visitor_id` attribute, but for remote assignments, [no value is returned](https://github.com/Betterment/test_track/blob/master/app/views/api/v1/visitors/_show.json.jbuilder#L2) in the response from the TT server.

So when you stuff a remote assignment into an `AnalyticsEvent`, you'll have a nil `visitor_id`, which will lead to errors.

I think the root cause is that `TestTrack::Remote::Assignment` has had an attribute for `visitor_id` for a long time, but it looks like it was never populated. So when [refactoring the analytics code](https://github.com/Betterment/test_track_rails_client/pull/44), we leaned on the existence of that attribute.

The fix is to pass in a `visitor_id` to the `AnalyticsEvent` and no longer delegate to the `Assignment`. I've also removed the visitor_id attributes from all the relevant classes to avoid confusion in the future.

### DIFF
--- a/app/models/test_track/offline_session.rb
+++ b/app/models/test_track/offline_session.rb
@@ -1,14 +1,4 @@
 class TestTrack::OfflineSession
-  AnalyticsAssignment = Struct.new(:visitor_id, :split_name, :variant, :context, :unsynced) do
-    def unsynced?
-      unsynced
-    end
-
-    def feature_gate?
-      split_name.end_with?('_enabled')
-    end
-  end
-
   def initialize(remote_visitor)
     @remote_visitor = remote_visitor
   end
@@ -40,20 +30,8 @@ class TestTrack::OfflineSession
   def visitor
     @visitor ||= TestTrack::Visitor.new(
       id: remote_visitor.id,
-      assignments: analytics_assignments
+      assignments: remote_visitor.assignments
     )
-  end
-
-  def analytics_assignments
-    remote_visitor.assignments.map do |remote_assignment|
-      AnalyticsAssignment.new(
-        remote_visitor.id,
-        remote_assignment.split_name,
-        remote_assignment.variant,
-        remote_assignment.context,
-        remote_assignment.unsynced
-      )
-    end
   end
 
   def manage
--- a/app/models/test_track/remote/assignment.rb
+++ b/app/models/test_track/remote/assignment.rb
@@ -9,6 +9,10 @@ class TestTrack::Remote::Assignment
     unsynced || variant_changed?
   end
 
+  def feature_gate?
+    split_name.end_with?('_enabled')
+  end
+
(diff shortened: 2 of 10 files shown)


app/models/test_track/remote/assignment.rb
```diff
@@ -1,14 +1,18 @@
 class TestTrack::Remote::Assignment
   include TestTrack::RemoteModel
 
-  attributes :visitor_id, :split_name, :variant, :context, :unsynced
+  attributes :split_name, :variant, :context, :unsynced
 
-  validates :visitor_id, :split_name, :variant, :mixpanel_result, presence: true
+  validates :split_name, :variant, :mixpanel_result, presence: true
```
samandmoore: do these validations ever get invoked?

app/models/test_track/offline_session.rb
```diff
@@ -1,14 +1,4 @@
 class TestTrack::OfflineSession
-  AnalyticsAssignment = Struct.new(:visitor_id, :split_name, :variant, :context, :unsynced) do
-    def unsynced?
-      unsynced
-    end
-
-    def feature_gate?
-      split_name.end_with?('_enabled')
-    end
-  end
```
samandmoore: so, remind me again why we needed this class? i swear we wouldn't have added it if it weren't necessary. we did some more refactoring so that we don't need it anymore?

samandmoore:
<<domain tafn -- just questions around the domain to make sure i'm following appropriately :)

app/models/test_track/remote/assignment.rb
aburgel: No because we never persist this model. But maybe they used to? Perhaps when we switched to using AssignmentEvent, we stopped persisting these.

app/models/test_track/offline_session.rb
aburgel: It was to get `TestTrack::Remote::Assignment` to look like `Assignment` when used by the analytics code. But there was another case that we missed... when unsynced assignments come back from the server in an online session.

We could wrap those instances in this AnalyticsAssignment wrapper, but I opted to change the contract of `TT::AnalyticsEvent` and move some of this functionality into `TT::Remote::Assignment` instead.

It feels like a win to have 2 Assignment-like classes rather than 3, but, tbh, I'm not sure if this was going somewhere else that I don't see.

aburgel:
bump

samandmoore:
Thanks for all the clarification

domainlgtm

PR #58, opened 2018-06-11
Sample experience events

jmileham:
/domain @samandmoore @aburgel 

Ok, here's the client side lib changes that pull in the new v2 split_registry and then respect the `experience_sampling_weight` property when sending analytics events.

I haven't tested this in an app yet, but I will before merging. The test suites pass. The second commit is big because it's a big hop to take that much API shape change and get the tests passing, but it's reasonably mechanical.

We continue to publish a v1-style split registry in the HTML for the JS client. At some point that'll want to change, but not til the JS client supports v2 split registries too. That'll be when we want to consider GZipping too.


app/controllers/tt/api/v1/split_registries_controller.rb
```diff
@@ -1,5 +0,0 @@
-class Tt::Api::V1::SplitRegistriesController < Tt::Api::V1::ApplicationController
-  def show
-    @active_splits = TestTrack::FakeServer.split_registry
-  end
-end
```
samandmoore: Do we wanna keep this for local dev including the js client until the js client is updated?

app/models/test_track/vary_dsl.rb
```diff
@@ -34,7 +34,7 @@ def default(variant, &block)
   delegate :split_name, to: :assignment
 
   def split
-    split_registry && split_registry[split_name]
+    split_registry && split_registry['splits'][split_name] && split_registry['splits'][split_name]['weights']
```
samandmoore: This nested hash lookup thing happening in several places is setting off my spideysense. It feels like we’re developing some object envy. 

Like maybe it would be nicer if split_registry responded to weights_for_split(split_name) or some sort of thing like that to provide some encapsulation?

samandmoore:
<<domain tafn

app/controllers/tt/api/v1/split_registries_controller.rb
jmileham: Yeah probs eh?

app/models/test_track/notify_assignment_job.rb
```diff
@@ -25,9 +25,30 @@ def perform
...
+  def skip_experience_event?
+    skip_all_experience_events? || !sample_event?
+  end
+
+  def skip_all_experience_events?
+    experience_sampling_weight.zero?
+  end
+
+  def experience_sampling_weight
+    @experience_sampling_weight ||= TestTrack::Remote::SplitRegistry.to_hash['experience_sampling_weight']
```
aburgel: to @samandmoore's point on object envy, i think this is another case where an object would provide a nicer interface.

aburgel:
<< platform

what @samandmoore said. tafn!

app/models/test_track/notify_assignment_job.rb
jmileham: Yeah, `SplitRegistry.experience_sampling_weight` is probably the right place to put that method unless we want to uncork a big refactor. No new object needed.

app/controllers/tt/api/v1/split_registries_controller.rb
jmileham: Turns out this was actually still conforming to the v1 interface :-/ I gotta fix it and support both. Though honestly, I'm not sure this endpoint is ever used by the JS client - the JS client only becomes aware of the TT server's URL from the split payload embedded in the page, with the split registry. Maybe I can kill them altogether instead. Will look into it.

aburgel: hmm, good point. the js client always gets the split registry from the page.

jmileham: pulled it out.

app/models/test_track/vary_dsl.rb
jmileham: wrapped it up into a global SplitRegistry cache.

app/models/test_track/notify_assignment_job.rb
jmileham: done

jmileham:
I have mixed feelings about how the global SplitRegistry cache played out, but it doesn't appear to be materially slower despite pulling the levers of rails caching for each hit to the split registry. It simplified the responsibilities of some things that were passing around split registries otherwise.

It also had a nice effect where if the TT server fails to respond to a given split_registry request, we now cache the absence for 5 seconds and muddle through rather than DDoSing the TT server.

app/models/test_track/remote/split_registry.rb
```diff
@@ -3,35 +3,61 @@ class TestTrack::Remote::SplitRegistry
...
+
+    def to_hash
+      if faked?
+        instance.attributes.freeze
+      else
+        fetch_cache { instance.attributes }.freeze
+      end
+    rescue *TestTrack::SERVER_ERRORS => e
+      Rails.logger.error "TestTrack failed to load split registry. #{e}"
+      fetch_cache { nil } # cache the missing registry for 5 seconds if we can't get one
```
aburgel: 👍

app/models/test_track/visitor.rb
```diff
@@ -62,10 +62,6 @@ def assignment_json
     end
   end
 
-  def split_registry
-    @split_registry ||= TestTrack::Remote::SplitRegistry.to_hash
```
aburgel: should we keep this and its usages above? it makes the vary code deterministic because its not dependent on the split registry being consistent and in the cache or available from the remote server.

i'm thinking about the (admittedly unlikely) case where the cache gets invalidated while running the vary algo, and either the weights change or the split registry goes offline.

jmileham: yeah, i mean, at scale unlikely happens a lot. problem is that this is a hash, and that's directly at odds with the object envy issue without a major refactor to be passing around a `SplitRegistry` instance of some sort that would implement a bunch of new methods instead of keeping this hash ref. So I see three paths:

* Accept this race condition - kinda ugly semantically
* revert, maybe keep the caching of nil which I like and eat the object envy for now
* Revert and go once more into the breach and take everything that used to hold reference to a registry hash to hold an instance of a nice object.

I'm sort of angling for option 2.

jmileham:
bump - i think this is good to go, pending some runs of real test suites with this version integrated

spec/models/test_track/notify_assignment_job_spec.rb
```diff
@@ -17,6 +17,25 @@
...
+              back_in_time: 100,
+              power_of_love: 0
+            },
+            feature_gate: false
+          }
+        },
+        experience_sampling_weight: 1
+      }.to_json)
+    example.run
+  end
```
aburgel: can we change this to stub `TestTrack::Remote::SplitRegistry.experience_sampling_weight`? then I think we can avoid stubbing `subject` on line 76 and 118.

aburgel:
>> platform
<< domain lgtm!

aburgel:
@aburgel >> platform

spec/models/test_track/notify_assignment_job_spec.rb
jmileham: k, sounds good!

jmileham: really can just stub splitregistry directly on those two lines and leave it "live" here, which works nice. did that.

aburgel:
should we land this? @jmileham @samandmoore

jmileham:
I’m encountering some test leakage in practice. Need to get to the bottom
of it. Will do this week hopefully.

aburgel:
Superseded by #82

PR #59, opened 2018-07-25
remove requirement of `context` to test track ab and vary dsl

effron:
/no-platform

Make the "context" argument optional when creating/fetching assignments using `Visitor#vary` and `Visitor#ab`. In practice, we don't always use context, and it makes using TestTrack in certain contexts difficult (we often have to make some meaningless context up just to satisfy the required parameter). 

Another approach would be to make this configurable per TestTrack::Identity -- the identity could set up whether or not its consumers require a `context`, and provide a default `context` if it isn't passed in. This only feels necessary if there are consumers that still want `context` to be required, and do not want there to ever be a `nil` context associated to an assignment

### DIFF
--- a/lib/test_track_rails_client/version.rb
+++ b/lib/test_track_rails_client/version.rb
@@ -1,3 +1,3 @@
 module TestTrackRailsClient
-  VERSION = "4.0.0.alpha8" # rubocop:disable Style/MutableConstant
+  VERSION = "4.0.0.alpha9" # rubocop:disable Style/MutableConstant
 end


app/models/test_track/remote/assignment_event.rb
```diff
@@ -3,7 +3,7 @@ class TestTrack::Remote::AssignmentEvent
 
   collection_path '/api/v1/assignment_event'
 
-  attributes :visitor_id, :split_name, :unsynced
+  attributes :visitor_id, :split_name, :unsynced, :context
```
effron: `context` is accepted on the other side of the wire, felt like it was worth including here for clarity

app/models/test_track/remote/assignment_event.rb
```diff
@@ -3,12 +3,10 @@ class TestTrack::Remote::AssignmentEvent
 
   collection_path '/api/v1/assignment_event'
 
-  attributes :visitor_id, :split_name, :unsynced
+  attributes :visitor_id, :split_name, :mixpanel_result, :context
```
effron: context is accepted on the other side of the wire (https://github.com/Betterment/test_track/blob/master/app/controllers/api/v1/assignment_events_controller.rb#L12), felt like it was worth including here for clarity

effron: also removing the unused `unsynced` attribute

jmileham: are you sure it's unused? Here's a candidate, and there's duck typing all over the place in this code.

https://github.com/Betterment/test_track_rails_client/blob/14b5650c2090635b6821462705af5f6efe51bd49/app/models/test_track/visitor.rb#L56

effron: it's used in assignments, but this class is only used for creating an AssignmentEvent, and the api in TestTrack only exposes a `create` endpoint (so this data isn't being read from anywhere that i can see)

jmileham: cool. i just am gunshy about removing things because the coupling is higher than i'd like at this point in the client code.

PR #61, opened 2018-07-27
remove requirement of `context` to test track ab and vary dsl

effron:
/no-platform

Make the "context" argument optional when creating/fetching assignments using Visitor#vary and Visitor#ab. In practice, we don't always use context, and it makes using TestTrack in certain contexts difficult (we often have to make some meaningless context up just to satisfy the required parameter).

Another approach would be to make this configurable per TestTrack::Identity -- the identity could set up whether or not its consumers require a context, and provide a default context if it isn't passed in. This only feels necessary if there are consumers that still want context to be required, and do not want there to ever be a nil context associated to an assignment

(reopening this https://github.com/Betterment/test_track_rails_client/pull/59 after i reset the master branch of my fork)

### DIFF
--- a/app/models/test_track/remote/assignment_event.rb
+++ b/app/models/test_track/remote/assignment_event.rb
@@ -3,7 +3,7 @@ class TestTrack::Remote::AssignmentEvent
 
   collection_path '/api/v1/assignment_event'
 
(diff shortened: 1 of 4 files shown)


effron:
Going to close this and do more research to see if we can remove `context` altogether instead of just making it optinal

PR #63, opened 2018-10-29
reraise all TestTrack server errors as unrecoverable connectivity errors

effron:
/domain @jmileham @aburgel 
/no-platform

A follow up to this (https://github.com/Betterment/test_track_rails_client/pull/62), this catches more common instances of visitor fetching that could result in a test track server error. In our use case, this impacts background jobs, especially mailers, so this would help us categorize these issues (which resolve themselves due to job retrys) and keep them out of our error tracker.

### DIFF
--- a/spec/models/concerns/test_track/remote_model_spec.rb
+++ b/spec/models/concerns/test_track/remote_model_spec.rb
@@ -1,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe TestTrack::Identity do
+  context 'handling server timeouts' do
+    describe 'for get requests' do
+      before do
+        stub_request(:get, 'http://testtrack.dev/api/v1/remotes/fake_id').to_timeout
+      end
+
+      it 'reraises them as TestTrack::UnrecoverableConnectivityError' do
+        with_test_track_enabled do
+          expect { Remote.find('fake_id') }.to raise_error(TestTrack::UnrecoverableConnectivityError)
+        end
+      end
+    end
+
+    describe 'for post requests' do
+      before do
+        stub_request(:post, 'http://testtrack.dev/api/v1/remotes').to_timeout
+      end
+
+      it 'reraises them as TestTrack::UnrecoverableConnectivityError' do
+        with_test_track_enabled do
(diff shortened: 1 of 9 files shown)


effron:
nanda?

jmileham:
should this not go into the guts of `from_identifier`?

jmileham:
<< domain

jmileham:
tafn

jmileham:
nanda

effron:
bump!

I refactored both of the places where UnrecoverableConnectivityErrors were being thrown into a middleware for the TestTrack::RemoteModel. All interactions with the TestTrack server will get this behavior by default, now.

jmileham:
So this seemed like a big hammer to me initially, but Harry and I talked about it offline. The takeaway is that TT client's software quality problems are not your (an app maintainer's) software quality problems. Even if we wrote a bug that hit the wrong route altogether and never worked, it still wouldn't be appropriate to raise an issue in the customer's Sentry. It'd be TT-core-team's fault for shipping such a wildly broken piece of software. So by that logic it seems right to give a big undifferentiated handle for ignoring all network connectivity issues from TT client, because they're all equally not your problem as a consumer, and practically speaking, they'll all be operational issues.

Also in terms of onboarding and maturing your TT integration, consumers will generally start not ignoring these exceptions, so if there were configuration issues leading to connectivity problems, they'd alert as software quality problems in the client app, as is appropriate. After they fix those, the only alerts they get will be occasional connectivity problems. When they hit certain scale, they will get annoying, and the obvious thing to do is ignore them as a class because the root cause is not a quality problem.

domain lgtm!

spec/models/concerns/test_track/remote_model_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+RSpec.describe TestTrack::Identity do
```
samandmoore: Is this supposed to be describing something else?

effron: yes, sorry about that. will pr a fix

PR #66, opened 2018-10-30
Add Ruby 2.4.2 to build matrix

### LINKED ISSUE #37: test against MRI 2.4.2
update the .travis.yml

smudge:
/domain @samandmoore @jmileham @effron
/platform @samandmoore @jmileham @effron

This fixes #37

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 
 rvm:
 - 2.3.1
+- 2.4.2
 - jruby-9.1.13.0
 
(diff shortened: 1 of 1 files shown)


jmileham:
<< domain platform

gonna wait on this build too. We can pull jruby out soon for performance wins.

jmileham:
(fwiw I'm not worried about this change not being in alphaN because tests don't ship in gem bundles anyway, so lets land this second.)

PR #67, opened 2018-11-09
Allow session cookie handling to work with localhost

aburgel:
/domain @jmileham @samandmoore @smudge @effron 
/no-platform

PublicSuffix currently blows up with a `PublicSuffix::DomainInvalid` exception if you attempt to parse `localhost`. This makes it impossible to run an app with this gem via `rails s` (since that starts a server on `localhost:3000`). This modifies session cookie handling to fallback to `request.host` if PublicSuffix is not able to parse the host.

It seems like we are intentionally relying on PublicSuffix to blow up to protect against invalid host names. I'm not sure we need this gem to provide that sort of security feature, so I'm removing it (since it gets in the way of this fix). But please push back if I'm missing something.

### DIFF
--- a/app/models/test_track/session.rb
+++ b/app/models/test_track/session.rb
@@ -95,10 +95,8 @@ class TestTrack::Session
   end
 
   def _cookie_domain
-    if bare_ip_address?
+    if bare_ip_address? || fully_qualified_cookie_domain_enabled?
       request.host
-    elsif fully_qualified_cookie_domain_enabled?
-      fully_qualified_domain
     else
       wildcard_domain
     end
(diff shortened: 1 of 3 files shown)


samandmoore:
i don't quite follow:

> It seems like we are intentionally relying on PublicSuffix to blow up to protect against invalid host names. I'm not sure we need this gem to provide that sort of security feature, so I'm removing it (since it gets in the way of this fix)

you _are_ removing it, or you _are not_ removing it?

i think @jmileham might have more valuable input on this PR than me. i would just like to see this work properly locally and also when the localhost is ipv6 which i've seen error with public suffix in other places before.

jmileham:
(feel free to bump version and release)

aburgel:
Added an issue to replace PublicSuffix with some Rails builtins now that we no longer depend on its security properties https://github.com/Betterment/test_track_rails_client/issues/68

aburgel:
> you are removing it, or you are not removing it?

My bad. I meant I'm leaving the gem, but removing the bit of code that blows up on a bad Host header.

> i would just like to see this work properly locally and also when the localhost is ipv6 which i've seen error with public suffix in other places before.

I think we correctly account for ipv6 addresses, since the regex we use to detect an ip also matches on ipv6. I haven't tested this though.

PR #69, opened 2019-02-03
add require_feature_flag class method to test track controller mixin

klesse413:
/domain @jmileham 
/no-platform

Adding `require_feature_flag` class method that we've previously defined in test track helpers to the controller mixin


klesse413:
woops sorry nanda

app/controllers/concerns/test_track/controller.rb
```diff
@@ -9,8 +9,20 @@ module TestTrack::Controller
...
+    def require_feature_flag(feature_flag, *args)
+      before_action(*args) do
+        not_found unless test_track_visitor.ab(feature_flag, context: self.class.name.demodulize.underscore)
+      end
+    end
+  end
+
   private
 
+  def not_found
```
jmileham: This is pretty a global name for a library mixin - we might conflict with an app developer’s different method. I’d probably roll this line into the callback

jmileham:
<< domain lgtm after that

jmileham:
Side effect of fork PR; if it works locally we’re good.

jmileham:
Feel free to bump minor too

jmileham:
Actually 4.2 seems borked. I dropped support for it, rails 5.0 and JRuby. Since we're getting ready for 4.0 and all, might as well make bold steps. :)

jmileham:
4.0.0.alpha15 is up on rubygems @klesse413 thanks!

PR #71, opened 2019-03-07
Fix migrations with multiple splits

smudge:
/domain @aburgel @Betterment/test_track_core @chukkwagon
/platform @aburgel
/cc @rahobbs

I tried un-memoizing, but that led to even more complexity, so I'm leaning in a bit more and memoize the splits that aren't on the server _before_ we start adding additional local splits.


test_track_rails_client.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'faraday_middleware'
   s.add_dependency 'mixpanel-ruby', '~> 1.4'
   s.add_dependency 'public_suffix', '>= 2.0.0', '<= 3.0.0'
-  s.add_dependency 'rails', '>= 4.1', "< 6.0"
+  s.add_dependency 'rails', '<= 5.2', '> 4.1'
```
smudge: `< 6.0` resolves to `6.0.0.beta1`, so I flipped it to `<= 5.2`

aburgel:
<< platformlgtm

nice fix!

chukkwagon:
<<Domain LGTM

PR #72, opened 2019-05-04
Auto prefix split names in vary and ab

jmileham:
/domain @smudge @samandmoore 
/platform @smudge @samandmoore 

This will allow you to use vary and ab calls with unprefixed split names that match your current app name, which is what will happen as soon as CLI-created splits are in play.


samandmoore:
Is this really all the places that need to change? Awesome

<<domainlgtm platformlgtm

PR #74, opened 2019-05-05
Testtrack cli migrations

jmileham:
/domain @samandmoore @smudge 
/platform @samandmoore @smudge 

To make deployment easy, I'm vendoring the testtrack CLI binaries (thankfully go is static) into the rails client, detecting the OS, and calling the appropriate one to `migrate`/`schema load` when apps call `rake db:migrate` and `rake db:schema:load`

We disable testtrack migrate/schema load in development, where people should only have fake servers.

This will not impact users in the field, as if you don't have a new-style testtrack schema, we'll fall back to old style schema loading and noop in `db:migrate`.

I don't have automated tests because orchestrating them would be tricky and be full of stubs which easily be incorrect and prove nothing about actual behavior. I acceptance tested the heck out of it, and it should be low-rate-of-change, so regression protection doesn't feel that important.

### DIFF
--- a/app/models/testtrack_cli.rb
+++ b/app/models/testtrack_cli.rb
@@ -1,0 +1,35 @@
+require 'rbconfig'
+
+class TesttrackCli
+  include Singleton
+
+  def project_initialized?
+    File.exist?(File.join('testtrack', 'schema.yml'))
+  end
+
+  def call(*args)
+    system(path, *args)
+    $CHILD_STATUS
+  end
+
+  def path
+    TestTrackRailsClient::Engine.root.join("vendor", "bin", "testtrack-cli", filename).to_s
+  end
+
+  def filename
+    case host_os
+      when /darwin/
+        "testtrack.darwin"
+      when /linux/
+        "testtrack.linux"
+      else
+        raise "no testtrack binary for platform #{host_os}"
+    end
+  end
+
+  private
+
+  def host_os
+    RbConfig::CONFIG['host_os']
+  end
+end
--- a/lib/tasks/test_track_rails_client_tasks.rake
+++ b/lib/tasks/test_track_rails_client_tasks.rake
@@ -26,7 +26,7 @@ namespace :test_track do
   end
 end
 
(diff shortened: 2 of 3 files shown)


lib/tasks/test_track_rails_client_tasks.rake
```diff
@@ -1,15 +1,33 @@
...
+        TestTrack.update_config do |c|
+          c.load_schema # Load legacy schema
+        end
       end
     end
   end
 end
 
-unless Rails.env.test?
+unless Rails.env.test? || (Rails.env.development? || !TesttrackCLI.instance.project_initialized?)
```
smudge: 👍

app/models/testtrack_cli.rb
```diff
@@ -0,0 +1,35 @@
+require 'rbconfig'
+
+class TesttrackCli
+  include Singleton
+
+  def project_initialized?
+    File.exist?(File.join('testtrack', 'schema.yml'))
```
smudge: Should this look at `Rails.root`?

smudge:
<< domain LGTM - just the one q

app/models/testtrack_cli.rb
jmileham: I was gonna have it do that but then I’d need to pushd into it to run tt cli, I futzed with that for a while and it was getting annoying (ruby stdlib Shell was logging nastiness out to the screen), so I figured this is symmetrical, and requires none of that.

samandmoore: I don't follow. Why not do Rails.root.join here?

samandmoore:
<<platformlgtm too

app/models/testtrack_cli.rb
jmileham: Because ultimately if the user isn’t in the app root, testtrack cli is going to fail, so unless I wire it up to work when calling rake from a random non-app-root directory, I might as well not app-root-bind the schema lookup. It’s gonna have a bad time if it thinks it’s setup and then can’t migrate or schema load

PR #76, opened 2019-05-05
Unroll method because referenced before class loading

jmileham:
/domain @samandmoore 
/platform @samandmoore 

Ok, this time?


samandmoore:
<<domainlgtm platformlgtm

NOT “test” OR NOT “it is development and new schema is set up”

PR #77, opened 2019-05-06
Only try to open files that exist

jmileham:
/domain @samandmoore @HipsterBrown
/platform @samandmoore 

I was relying on YAML.load_file to work like the previous code assumed it did (and had tests that made that assumption explicit). I was skeptical of this behavior at the time, but didn't do the research. :)

So here we go, better than ever - this used to blow up if you didn't have an old-school test track schema file too, instead of returning empty bags like the tests indicated.


samandmoore:
<<domainlgtm platformlgtm

PR #79, opened 2019-05-09
Alpha25 - CLI 0.9.5 makes retirement migrations idempotent

jmileham:
This allows you to change the decision of a retired split without having
to revive the split first.

/domain @samandmoore 
/no-platform

We're gaining back those KBs on the binaries with all these fancy features


samandmoore:
<<domainlgtm

PR #80, opened 2019-05-20
Stub prefixed split names in stub_test_track_assignments

danf1024:
/domain @Betterment/test_track_core @effron 
/platform @smudge @jmileham @aburgel 

now that `vary` and `ab` auto-prefix split names (https://github.com/Betterment/test_track_rails_client/pull/72) it follows that `stub_test_track_assignments` should also auto-prefix the split names it stubs---so that a developer can stub using the same split name they pass to `vary` and `ab`.

### DIFF
--- a/test_track_rails_client.gemspec
+++ b/test_track_rails_client.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "faraday", ">= 0.8", "< 1.0"
   s.add_dependency 'faraday_middleware'
   s.add_dependency 'mixpanel-ruby', '~> 1.4'
+  s.add_dependency 'multi_json', '~> 1.7'
   s.add_dependency 'public_suffix', '>= 2.0.0', '<= 3.0.0'
   s.add_dependency 'rails', '>= 4.1', "< 6.0"
   s.add_dependency 'request_store', '~> 1.3'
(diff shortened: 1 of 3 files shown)


danf1024:
nanda?

danf1024:
nanda?

jmileham:
Should we only create the prefixed assignment if the prefixed split exists in the fake split registry? I think probably that would help us detect typing errors. We have this cool property now where splits prefixed with our own app name are guaranteed to be in the schema file or else be invalid.

danf1024:
@jmileham check the latest push and lmk if it aligns with what you're saying

test_track_rails_client.gemspec
```diff
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.add_dependency "faraday", ">= 0.8", "< 1.0"
   s.add_dependency 'faraday_middleware'
   s.add_dependency 'mixpanel-ruby', '~> 1.4'
+  s.add_dependency 'multi_json', '~> 1.7'
```
jmileham: Is this where you landed? We need this to reliably support vendored Her gem if nothing else depends on it?

jmileham:
<< domain lgtm platform lgtm!

test_track_rails_client.gemspec
danf1024: yup...the same failure came up on CI before i added this, so it does seem like some transitive dependency disappeared, requiring us to now explicitly define it.

PR #65, opened 2018-10-30
Support Rails 5.2 (plus tests)

smudge:
/domain @samandmoore @jmileham @effron 
/platform @samandmoore @jmileham @effron 

This vendors a 5.2-compatible version of `her` and fixes an 5.2-specific breakage in `fakeable_her`.

The files in `gemfiles/` were all regenerated by `appraisal` - looks like we bumped `nokogiri` in `Gemfile` without bumping it in the Rails version builds.

Rails 5.2 is now included in the Travis build matrix.

PR #81, opened 2019-05-21
Add script for vending testtrack-cli

aburgel:
/domain @jmileham @smudge @samandmoore 
/no-platform

Added a step to the `vendor_deps` task to download the CLI to the correct directory (we had multiple versions committed in different directories).

Also, upgraded the CLI to v0.9.7.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -34,8 +34,17 @@ task :vendor_deps do
 
   # Download testtrack-cli
   TEST_TRACK_CLI_VERSION = 'v0.9.7'
-  %w(darwin linux).each do |arch|
-    `curl -L https://github.com/Betterment/testtrack-cli/releases/download/#{TEST_TRACK_CLI_VERSION}/testtrack.#{arch} > vendor/bin/testtrack-cli/testtrack.#{arch}`
+  FileUtils.module_eval do
+    mkdir_p 'vendor/bin/testtrack-cli'
+    cd 'vendor/bin/testtrack-cli' do
+      rm_r Dir.glob('*')
+
+      %w(darwin linux).each do |arch|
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -32,6 +32,21 @@ task :vendor_deps do
   sh 'bower install'
   sh 'cp', 'bower_components/test_track_js_client/dist/testTrack.bundle.min.js', 'app/assets/javascripts'
 
+  # Download testtrack-cli
+  TEST_TRACK_CLI_VERSION = 'v0.9.7'
```
smudge: So if we want to upgrade the vendored CLI version, do we just change this constant?

aburgel: yep, and then run `bundle exec rake vendor_deps`... there's a bit of noise w/ fakeable_her.  we are not running the latest for reasons that i don't understand (you actually addressed it in #65). so you need to not commit those changes.

jmileham: should we rip out the rest of the gem upgrades and let them rot?

Rakefile
```diff
@@ -32,6 +32,21 @@ task :vendor_deps do
...
+  FileUtils.module_eval do
+    mkdir_p 'vendor/bin/testtrack-cli'
+    cd 'vendor/bin/testtrack-cli' do
+      rm_r Dir.glob('*')
+
+      %w(darwin linux).each do |arch|
+        `curl -L https://github.com/Betterment/testtrack-cli/releases/download/#{TEST_TRACK_CLI_VERSION}/testtrack.#{arch} \
```
jmileham: One fun thing I learned is that this gets rate limited pretty readily when used unauthenticated. I spose it doesn't matter much for this operation, though, you only rarely need to do this, and can cope with some pain if it arises.

jmileham:
<< domain lgtm

Rakefile
aburgel: works for me, we haven't touched those gems in a long time. i can do this separately.

PR #82, opened 2019-05-23
Sample experience events

aburgel:
/no-platform

I've resurrected #58, rebased, and fixed a couple of small things. That PR had a single LGTM by me, and most of this code was written by @jmileham, so I'd like to get some other eyes on this (@samandmoore, @smudge?).

I have a [separate branch](https://github.com/Betterment/test_track_rails_client/compare/aburgel/sample_experience_events...aburgel/object_envy) that adds back in the SplitRegistry object that @jmileham had partially introduced. Since this PR is mostly the same as #58, I think it'll be easier to do another review on this first and quickly follow with the remaining work. But I'm happy to just open up one big PR if that's preferable to others.


lib/test_track_rails_client/assignment_helper.rb
```diff
@@ -2,16 +2,18 @@ module TestTrackRailsClient::AssignmentHelper
   def stub_test_track_assignments(assignment_registry) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
     raise "Cannot stub test track assignments when TestTrack is enabled" if TestTrack.enabled?
 
-    split_registry = TestTrack::Fake::SplitRegistry.instance.to_h.dup
+    split_registry = TestTrack::Fake::SplitRegistry.instance.to_h.deep_dup
```
jmileham: was this the leakage?

jmileham:
BTW, IIRC I ran into trouble with test leakage when integrating into an app, even when this test suite passed. Have you linked this up to an app and run the full suite to make sure that's all cool?

lib/test_track_rails_client/assignment_helper.rb
aburgel: yeah, this was the leakage. the inner `split` hash was not duped, so modifications to it was persisted across tests.

i'll run it against a few apps to see if there is anything else going on.

jmileham:
I'm happy to take this, It's not fresh enough that I feel any sense of ownership :-) or even understand it anymore.

<< domain

jmileham:
love it. if tests of a big client app pass, domain lgtm!

app/models/test_track/remote/split_registry.rb
```diff
@@ -3,35 +3,45 @@ class TestTrack::Remote::SplitRegistry
...
+
+    def to_hash
+      if faked?
+        instance.attributes.freeze
+      else
+        fetch_cache { instance.attributes }.freeze
+      end
+    rescue *TestTrack::SERVER_ERRORS => e
+      Rails.logger.error "TestTrack failed to load split registry. #{e}"
+      fetch_cache { nil } # cache the missing registry for 5 seconds if we can't get one
```
smudge: Should this return a `{}`? We recently saw a nil object error down from the `nil` return value in this method.

I'm also a fan of `to_x` methods returning an X no matter how nonsensical (and relying on input/value sanitization to handle failure states).

smudge:
Sweet 👍

app/models/test_track/remote/split_registry.rb
aburgel: for this PR there's a bunch of code that depends on this being nil, but in #83 i can have this return `{}`

aburgel:
ok, tested against a couple of apps and found 1 bug related to prefixing. its fixed so merging away!

PR #83, opened 2019-05-23
Introduce a SplitRegistry model to avoid passing around hashes

aburgel:
/no-platform

A follow on to #82 (and mostly based on @jmileham's work that he reverted in #58). ~~I'm keeping this in draft mode until #82 lands. I think I need to add a few more specs too.~~

TODO
- [x] Run against an app's test suite

### DIFF
--- a/app/models/test_track/split_registry.rb
+++ b/app/models/test_track/split_registry.rb
@@ -1,0 +1,39 @@
+class TestTrack::SplitRegistry
+  def self.from_remote
+    new(TestTrack::Remote::SplitRegistry.to_hash)
+  end
+
+  def initialize(registry_hash)
+    @registry_hash = registry_hash
+  end
+
+  def include?(split_name)
+    registry_hash['splits'].key?(split_name)
+  end
+
+  def loaded?
+    registry_hash.present?
+  end
+
+  def split_names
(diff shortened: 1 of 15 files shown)


app/models/test_track/split_registry.rb
```diff
@@ -0,0 +1,39 @@
...
+  def split_names
+    registry_hash['splits'].keys
+  end
+
+  def experience_sampling_weight
+    registry_hash.fetch('experience_sampling_weight')
+  end
+
+  def weights_for(split_name)
+    registry_hash && registry_hash['splits'][split_name] && registry_hash['splits'][split_name]['weights']
```
aburgel: TODO: freeze the weights hash to be a little safer

jmileham: like we could freeze the entire registry_hash on ingest, I figure, right?

jmileham: (deep freeze)

aburgel:
@Betterment/test_track_core this is ready for review

aburgel:
@jmileham ran this against a few apps and their test suites, all looks good. that mystery guest issue seems to have been resolved.

jmileham:
<< domain lgtm! Thank you!

Issue #55, opened 2018-05-09
Fully deprecate public use of `OfflineSession`s via managed RequestStore-based sessions in non-web contexts


jmileham:
Delayed::Job and Clockwork contexts should manage an offline session for you, allowing session lifecycle management/caching of an arbitrary number of identity-driven TestTrack visitors.

PR #84, opened 2019-05-27
Add a new session type for job queues

aburgel:
/no-platform

This introduces a `JobSession` (along with a DelayedJob plugin that makes use of it) that manages the lifecycle and keeps a cache of visitors. This improves performance by eliminating duplicate identity lookups within a job.

This addresses some of #55, but it keeps the OfflineSession for unmanaged contexts, like the command line and for looking up visitors by visitor ID.

I renamed the existing `Session` model to `WebSession` for clarity, this is much of the diff. There's a tiny bit of overlap between JobSession and SessionVisitorRepository, but it was only a few lines, so I opted for duplication rather than refactoring.


app/models/test_track/web_session.rb
```diff
@@ -1,7 +1,4 @@
-require 'delayed_job'
-require 'delayed_job_active_record'
```
aburgel: I'm not sure why we had these here. We require `delayed_job` in the engine, so this seems unnecessary. I'll see if any apps blow up with this change before merging.

jmileham:
Cool, if this doesn't cause any test leakage or other issues with the require statements you pulled << domain lgtm! Thanks so much!

PR #85, opened 2019-06-18
remove hard dep on airbrake in favor of optional one

samandmoore:
this changeset makes it so that we only interact with Airbrake if
there's an airbrake constant loaded. we also load the airbrake dj plugin
if the gem has been specified

/domain @aburgel @jmileham @smudge 
/no-platform

so, i'm opening this as a draft PR. ideally, we'd extend TT rails client to allow the end-user to configure their own misconfiguration notifier and default to the Airbrake one if the Airbrake gem is present? And then we'd set up a Null notifier when it's not?


samandmoore:
added a second commit to make misconfiguration notifier configurable https://github.com/Betterment/test_track_rails_client/pull/85/commits/e17200e815af2a27aea207c74d85e294930ec0f7

jmileham:
with that second commit, do you want to remove the draftiness of this PR?

app/models/test_track/misconfiguration_notifier.rb
```diff
@@ -1,13 +1,33 @@
...
+    def notify(_); end
+  end
 
-    if Airbrake.respond_to?(:notify_or_ignore)
-      Airbrake.notify_or_ignore(StandardError.new, error_message: msg)
-    else
-      Airbrake.notify(StandardError.new, error_message: msg)
+  class Airbrake
+    def notify(msg)
+      if defined?(::Airbrake)
```
jmileham: should this switch to a raisey guard clause? At this point you will have configured the Airbrake notifier, so if it isn't there we got problems, right?

lib/test_track.rb
```diff
@@ -21,16 +21,32 @@ module TestTrack
...
-      @analytics = client.is_a?(TestTrack::Analytics::SafeWrapper) ? client : wrapper(client)
+      @analytics = client.is_a?(TestTrack::Analytics::SafeWrapper) ? client : analytics_wrapper(client)
+    end
+
+    def misconfiguration_notifier
+      @misconfiguration_notifier ||= MisconfigurationNotifier::Wrapper.new(default_notifier)
+    end
+
+    def misconfiguration_notifier=(notifier)
+      @misconfiguration_notifier = MisconfigurationNotifier::Wrapper.new(notifier)
```
jmileham: Is the only kind of explicit Errbit-style notifying we do in MisconfigurationNotifier? Everything else we just figure you've wired in an exception alerting tool somewhere up stack?

spec/models/test_track/misconfiguration_notifier_spec.rb
```diff
@@ -37,31 +37,35 @@
       end
 
       context "given an Airbrake that responds to .notify_or_ignore and .notify" do
+        subject { TestTrack::MisconfigurationNotifier::Wrapper.new(TestTrack::MisconfigurationNotifier::Airbrake.new) }
+
         before do
           stub_const("Airbrake", double("Airbrake", notify: nil, notify_or_ignore: nil))
```
jmileham: There should be a test where Airbrake isn't defined for both the TestTrack.misconfiguration_notifier lazy getter and for the notifier itself, right?

jmileham:
<< domain tafn. Thanks for doing this!

samandmoore:
> with that second commit, do you want to remove the draftiness of this PR?

@jmileham sure, if y'all are into this idea, otherwise imma just close it :)

jmileham:
soooper into it.

jmileham:
also prolly needs a note in the README

lib/test_track.rb
samandmoore: Yea, grepping for Airbrake only turned up this file and its specs

spec/models/test_track/misconfiguration_notifier_spec.rb
samandmoore: hm, i'm not sure i follow. is that covered in the test on line 34? these tests are about the MisconfiguratonNotifier class and its behavior and not the TestTrack.misconfiguration_notifier behavior. i can add tests for that separately.

samandmoore:
bump

i gave the README a go.

spec/test_track_spec.rb
```diff
@@ -67,6 +67,38 @@
...
+        begin
+          default_notifier = TestTrack.misconfiguration_notifier
+          stub_const("Airbrake", double("Airbrake"))
+          TestTrack.remove_instance_variable(:@misconfiguration_notifier) if TestTrack.instance_variable_defined?(:@misconfiguration_notifier)
+          expect(TestTrack.misconfiguration_notifier.underlying.class).to eq TestTrack::MisconfigurationNotifier::Airbrake
+        ensure
+          TestTrack.misconfiguration_notifier = default_notifier
+        end
+      end
+    end
```
samandmoore: this test is kinda nasty, but such is life

README.md
```diff
@@ -369,6 +369,24 @@ mixpanel.init('YOUR MIXPANEL TOKEN', {
 });
 ```
 
+## Misconfiguration notifications
+
+TestTrack provides hooks to easily integrate with your preferred error catching tool to receive notifications when we detect a misconfiguration in split configuration. TestTrack has built-in support for `Airbrake`. By default, if you've included the `airbrake` gem, TestTrack will use `Airbrake` as a misconfiguration notifier backend. If you wish to use another tool, you can set the `misconfiguration_notifier` attribute on `TestTrack` with your custom client. You should do this in a Rails initializer.
```
samandmoore:
```suggestion
TestTrack provides hooks to easily integrate with your preferred error catching tool to receive notifications when we detect a misconfiguration in split usage. TestTrack has built-in support for `Airbrake`. By default, if you've included the `airbrake` gem, TestTrack will use `Airbrake` as a misconfiguration notifier backend. If you wish to use another tool, you can set the `misconfiguration_notifier` attribute on `TestTrack` with your custom client. You should do this in a Rails initializer.
```

app/models/test_track/misconfiguration_notifier.rb
```diff
@@ -1,13 +1,32 @@
...
+    def notify(_); end
+  end
 
-    if Airbrake.respond_to?(:notify_or_ignore)
-      Airbrake.notify_or_ignore(StandardError.new, error_message: msg)
-    else
-      Airbrake.notify(StandardError.new, error_message: msg)
+  class Airbrake
+    def notify(msg)
+      raise "Aibrake was configured not found" unless defined?(::Airbrake)
```
aburgel: `Aibrake was not found`? or is this just missing a comma?

samandmoore: oh, i think i need a `but`

samandmoore:
```suggestion
      raise "Aibrake was configured but not found" unless defined?(::Airbrake)
```

aburgel:
@jmileham >> domain
<< domain lgtm

spec/models/test_track/misconfiguration_notifier_spec.rb
jmileham: got it. line 34 wasn't visible in the diff, obviously, so I missed that that case was already covered. But you could do things like assert that Aibrake isn't defined or otherwise document that that's what's being exercised in the test title.

PR #87, opened 2019-07-12
Always stub split registry in stub_test_track_assignments

danf1024:
/no-platform

Tracked down some unexpected behavior with the stubbing of prefixed vs non-prefixed split names...

In the conditional where we'd append a prefix to a passed in un-prefixed split name, we were only stubbing visitor assignments, whereas the other side of the conditional would stub both visitor assignments and override weights in the split registry. It turns out the overriding of the split registry is important in contexts where there is no pre-existing visitor (e.g. an unauthenticated context which will generate a new visitor). In that case, assignments are not used (see https://github.com/Betterment/test_track_rails_client/blob/master/app/models/test_track/visitor.rb#L13) so overriding the split registry is the only means by which we can actually stub the generated assignment.

### DIFF
--- a/lib/test_track_rails_client/assignment_helper.rb
+++ b/lib/test_track_rails_client/assignment_helper.rb
@@ -8,10 +8,17 @@ module TestTrackRailsClient::AssignmentHelper
 
     assignment_registry.each do |split_name, variant|
       prefixed_split_name = "#{app_name}.#{split_name}"
-      split_name = prefixed_split_name if split_registry['splits'].key?(prefixed_split_name)
+      split_name = if split_registry['splits'].key?(prefixed_split_name)
+                     prefixed_split_name
+                   else
+                     split_name.to_s
+                   end
 
-      split_registry['splits'][split_name.to_s] = { weights: { variant.to_s => 100 }, feature_gate: false }
-      assignments << { split_name: split_name.to_s, variant: variant.to_s, unsynced: false }
+      split_registry['splits'][split_name] = {
+        weights: { variant.to_s => 100 },
+        feature_gate: split_name.end_with?('_enabled')
+      }
+      assignments << { split_name: split_name, variant: variant.to_s, unsynced: false }
     end
 
     visitor_attributes = { id: "fake_visitor_id", assignments: assignments }
--- a/lib/test_track_rails_client/assignment_helper.rb
+++ b/lib/test_track_rails_client/assignment_helper.rb
@@ -8,14 +8,10 @@ module TestTrackRailsClient::AssignmentHelper
 
     assignment_registry.each do |split_name, variant|
       prefixed_split_name = "#{app_name}.#{split_name}"
-      if split_registry['splits'].key?(prefixed_split_name)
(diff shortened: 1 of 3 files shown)


lib/test_track_rails_client/assignment_helper.rb
```diff
@@ -8,14 +8,10 @@ def stub_test_track_assignments(assignment_registry) # rubocop:disable Metrics/A
 
     assignment_registry.each do |split_name, variant|
       prefixed_split_name = "#{app_name}.#{split_name}"
-      if split_registry['splits'].key?(prefixed_split_name)
-        assignments << { split_name: prefixed_split_name, variant: variant.to_s, unsynced: false }
-      else
-        unless split_registry['splits'][split_name]
```
danf1024: The existence of this conditional seems dubious. I can't see why we wouldn't want this split registry override always (regardless of whether the split is already in the registry), because of the visitor-generation scenario I described in the description. 

It also created some unexpected behavior based on whether the split name was passed in as a string or a symbol. Such that if you were trying to stub for one of those visitor-generating contexts and you passed in the split name as a string instead of a symbol, the stub wouldn't actually be applied.

jmileham: This makes sense. I think it probably was just trying to be minimally invasive. For example, we're totally hardcoding the `feature_gate` property to false when we "make up" a split. We should maybe consider calculating that from suffix just to be more legit, if we're going to override the split registry in all cases?

jmileham:
Have you tested this in integration? Another thing I'd want to be sure doesn't happen is the client code yelling at us because the set of variants in the registry doesn't coincide with the set of variants in the vary call because this registry faking only drops the selected variant at 100%, no other variants at 0%.

<< domain tafn

This could be good to merge, just let me know your thoughts on those two questions.

jmileham:
(and thank you for digging in to find this! this is gnarly!)

lib/test_track_rails_client/assignment_helper.rb
danf1024: yup, i'll make that change.

danf1024:
@jmileham i had a successful ci run for integration this morning, going to run it again now with the commit i just pushed to be sure.

jmileham:
great!

domain lgtm! Let me know when you've got confirmation that everything's cool and I'll release.

Can you bump the minor version as well? This feels like a bug fix to me, though the addition of a hard-weighted registry entry is arguably behavior change, but it's only in test, and we've got reason to believe it won't impact end users even in tests.

lib/test_track_rails_client/version.rb
```diff
@@ -1,3 +1,3 @@
 module TestTrackRailsClient
-  VERSION = "4.0.0.alpha27" # rubocop:disable Style/MutableConstant
+  VERSION = "4.0.0.alpha28" # rubocop:disable Style/MutableConstant
```
jmileham: Forgot we're in an interminable alpha anyway :) Thanks!

danf1024:
@jmileham version bumped and good to go

PR #88, opened 2019-07-17
 Add an escape hatch for skipping TT CLI schema/migrate rake tasks

smudge:
/platform @jmileham 

I made the choice to pull the giant conditional out into a method on the singleton, but since `TesttrackCli` isn't loaded yet at the top level of the rake task, I pulled the escape hatch down into a guard clause in the actual task blocks.

### DIFF
--- a/app/models/testtrack_cli.rb
+++ b/app/models/testtrack_cli.rb
@@ -3,6 +3,12 @@ require 'rbconfig'
 class TesttrackCli
   include Singleton
 
+  def skip_testtrack_cli?
+    Rails.env.test? ||
+      (Rails.env.development? && !project_initialized?) ||
+      ENV['SKIP_TESTTRACK_CLI'].present?
+  end
+
   def project_initialized?
     File.exist?(File.join('testtrack', 'schema.yml'))
   end
--- a/app/models/testtrack_cli.rb
+++ b/app/models/testtrack_cli.rb
@@ -4,9 +4,7 @@ class TesttrackCli
   include Singleton
 
   def skip_testtrack_cli?
-    Rails.env.test? ||
-      (Rails.env.development? && !project_initialized?) ||
-      ENV['SKIP_TESTTRACK_CLI'].present?
+    Rails.env.test? || (Rails.env.development? && project_initialized?) || ENV['SKIP_TESTTRACK_CLI'].present?
   end
 
   def project_initialized?
--- a/app/models/testtrack_cli.rb
+++ b/app/models/testtrack_cli.rb
@@ -4,7 +4,7 @@ class TesttrackCli
   include Singleton
 
   def skip_testtrack_cli?
-    Rails.env.test? || (Rails.env.development? && project_initialized?) || ENV['SKIP_TESTTRACK_CLI'].present?
+    Rails.env.test? || (Rails.env.development? && project_initialized?) || ENV.key?('SKIP_TESTTRACK_CLI')
   end
 
   def project_initialized?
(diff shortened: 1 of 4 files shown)


jmileham:
Ok. I hurt myself multiple times moving these conditionals around. Have you tested manually in a variety of dev/prod/test scenarios?

jmileham:
(not that I've tried this combo, and it seems more robust/simple than what I'd attempted, because I was maybe needlessly hung up on not mixing in those rake tasks unless they were gonna be used, but just given the gnarly rollout picture of a buggy version of this code, it's worth being sure)

jmileham:
<< domain platform tafn

smudge:
Ok without expanding the entire truth table, and using a much more terse rspec syntax than we usually want, I enumerated a bunch of test cases and tested it both with the negated version of the original condition:

```ruby
subject { !(!Rails.env.test? && !(Rails.env.development? && TesttrackCli.instance.project_initialized?) && !ENV['SKIP_TESTTRACK_CLI') }
```

and with the new codepath:

```ruby
subject { TesttrackCli.instance.skip_testtrack_cli? }
```

Behavior seems the same, tests pass in both cases.

I also pointed my local app to this branch and ran `rake db:setup` in test and development envs, but I'm not sure about how to test this in prod since I cant RAILS_ENV=production with a working configuration locally. (Hence the rspec tests cases.)

smudge:
(bump)

app/models/testtrack_cli.rb
```diff
@@ -3,6 +3,10 @@
 class TesttrackCli
   include Singleton
 
+  def skip_testtrack_cli?
+    Rails.env.test? || (Rails.env.development? && project_initialized?) || ENV['SKIP_TESTTRACK_CLI'].present?
```
jmileham: nitting hard

```suggestion
    Rails.env.test? || (Rails.env.development? && project_initialized?) || ENV.key?('SKIP_TESTTRACK_CLI')
```

jmileham: should we run it through a bool-like-caster thingie so nobody gets caught out by setting `false` `f` or `0`?

jmileham:
Don't have super strong feels on bool casting, but looks great. Thanks for testing in integration. Once the rake tasks work at all without blowing up, the test coverage you dropped gives more than enough confidence.

<< domain lgtm platform lgtm!

app/models/testtrack_cli.rb
smudge: Yeah we could. It's easiest with activesupport which we should have here since Rails is a dep

spec/models/testtrack_cli_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.describe TesttrackCli do
+  describe '#skip_testtrack_cli?' do
+    subject { TesttrackCli.instance.skip_testtrack_cli? }
+
+    before do |example|
+      mock_env = example.metadata[:env] || 'production'
+      allow(Rails).to receive(:env) { mock_env.inquiry }
+      allow(ENV).to receive(:[]).with('SKIP_TESTTRACK_CLI') { '1' } if example.metadata[:env_skip]
```
jmileham:
```suggestion
      allow(ENV).to receive(:key?).with('SKIP_TESTTRACK_CLI') { true } if example.metadata[:env_skip]
```

jmileham:
Added a suggestion that might fix the specs. Also a version bump and rubygems release would be grand, whenever you get to it!

PR #89, opened 2019-07-22
Accept a class name, not an instance reference

jmileham:
Prevents class cache nonsense when used with spring by reifying a fresh
instance on each call.

In usage, this value is ephemerally memoized in VaryDSL and ABConfiguration instances, which are short lived, so we should be fine. Slight reduction in performance when we find a misconfigured split, but that seems fine steady state, especially since we've never profiled this code or optimized it to begin with.

I think this config option was introduced in an alpha, so this shouldn't require bumping version beyond yet-another-alpha. I think we're getting close to cutting an RC, finally.

/domain @samandmoore
/platform @samandmoore

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,10 +375,11 @@ TestTrack provides hooks to easily integrate with your preferred error catching 
 
 ```ruby
 # config/initializers/test_track.rb
-TestTrack.misconfiguration_notifier = MyCustomMisconfigurationNotifier.new
+TestTrack.misconfiguration_notifier_class_name = 'MyCustomMisconfigurationNotifier'
 ```
 
-Your client must implement the following methods:
+Your client must be a class that can be instantiated with no arguments,
+and implement the following methods:
 
 ```ruby
 # Called when a Split misconfiguration is detected
--- a/spec/test_track_spec.rb
+++ b/spec/test_track_spec.rb
@@ -68,36 +68,51 @@ RSpec.describe TestTrack do
   end
 
   describe "misconfguration_notifier" do
-    it "wraps a custom notifier in Wrapper" do
+    it "wraps a custom notifier in Wrapper without memoizing" do
       begin
-        default_notifier = TestTrack.misconfiguration_notifier
         fake_notifier = double
-        TestTrack.misconfiguration_notifier = fake_notifier
-        expect(TestTrack.misconfiguration_notifier.class).to eq TestTrack::MisconfigurationNotifier::Wrapper
+        fake_notifier_class = double(new: fake_notifier)
+        stub_const('FakeNotifier', fake_notifier_class)
+        TestTrack.misconfiguration_notifier_class_name = 'FakeNotifier'
+
+        expect(TestTrack.misconfiguration_notifier).to be_instance_of(TestTrack::MisconfigurationNotifier::Wrapper)
         expect(TestTrack.misconfiguration_notifier.underlying).to eq fake_notifier
       ensure
-        TestTrack.misconfiguration_notifier = default_notifier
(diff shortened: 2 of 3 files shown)


spec/test_track_spec.rb
```diff
@@ -68,37 +68,52 @@
...
-      expect(TestTrack.misconfiguration_notifier.underlying.class).to eq TestTrack::MisconfigurationNotifier::Null
+    it "returns a new instance each time" do
+      begin
+        fake_notifier = double
+        fake_notifier_class = double
+        call_count = 0
+        allow(fake_notifier_class).to receive(:new) do
+          call_count += 1
+          fake_notifier
+        end
```
samandmoore: we can't use the built-in rspec call count stuff for this?

jmileham: The validation calls it once, and I don't want to write a spec that says that two calls will result in 3 calls to the underlying instantiation.

samandmoore:
<<domainlgtm platformlgtm

i might change the spec to use the built-in rspec `expect(blah).to have_received(:new).exactly(3).times` or whatever

jmileham:
See comment above why not ^^ :)

jmileham:
nanda - I made major changes to bring the analytics setter over to this pattern as well and invalidated the previous signoff received. Holler.

lib/test_track_rails_client/version.rb
```diff
@@ -1,3 +1,3 @@
 module TestTrackRailsClient
-  VERSION = "4.0.0.alpha28" # rubocop:disable Style/MutableConstant
+  VERSION = "4.0.0.alpha29" # rubocop:disable Style/MutableConstant
```
jmileham:
```suggestion
  VERSION = "4.0.0.alpha30" # rubocop:disable Style/MutableConstant
```

samandmoore:
still looks good to me

README.md
```diff
@@ -336,10 +336,10 @@ TestTrack does not offer built-in functionality for analyzing the results of spl
 
 ```ruby
 # config/initializers/test_track.rb
-TestTrack.analytics = MyCustomAnalyticsClient.new
+TestTrack.analytics_class_name = 'MyCustomAnalyticsClient'
 ```
 
-Your client must implement the following methods:
+Your client must be a singleton or requre no initializer arguments, implement the following methods:
```
aburgel:
```suggestion
Your client must be a singleton or require no initializer arguments, implement the following methods:
```

README.md
```diff
@@ -375,10 +375,10 @@ TestTrack provides hooks to easily integrate with your preferred error catching
 
 ```ruby
 # config/initializers/test_track.rb
-TestTrack.misconfiguration_notifier = MyCustomMisconfigurationNotifier.new
+TestTrack.misconfiguration_notifier_class_name = 'MyCustomMisconfigurationNotifier'
 ```
 
-Your client must implement the following methods:
+Your client must be a singleton or requre no initializer arguments, implement the following methods:
```
aburgel:
```suggestion
Your client must be a singleton or require no initializer arguments, implement the following methods:
```

jmileham:
cool. looking for a final signoff here nanda. I know I invalidated your prior LG Sam. @aburgel it looks like you've been a-looking, if you wanna steal Sam's claims and give that signoff, that would be chill.

samandmoore:
@samandmoore >> domain
@samandmoore >> platform
@aburgel << domain platform

aburgel:
domainlgtm platformlgtm

jmileham:
So this is breaking, right, so we should be sure to merge before declaring 4.0 final. I can get to it Wednesday if you don't first @aburgel. :)

PR #91, opened 2019-08-27
Allow job session to be reentrant

aburgel:
/domain @jmileham @smudge @samandmoore 
/no-platform

Further testing of #84 revealed that running DJ in synchronous mode (during tests) causes plugins to be executed on each invocation. So if you have a job the runs another job, you'll have plugins running within plugins. So the extra bit of safety I added to prevent misuse is too restrictive.

I've modified the job plugin to allow reentrancy. If this were to occur in production, you'd lose some efficiency of visitor caching, but you would not lose correctness.

### DIFF
--- a/app/models/test_track/job_session.rb
+++ b/app/models/test_track/job_session.rb
@@ -1,7 +1,6 @@
 class TestTrack::JobSession
   def manage
     raise ArgumentError, "must provide block to `manage`" unless block_given?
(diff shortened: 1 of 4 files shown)


jmileham:
<< domain lgtm

Yeah, job inlining would be incredibly rare in production - if you did want to call a job, you'd just call perform on a job object, and none of this would apply, so this seems like it will cost nothing in practice and be good for test use.

PR #92, opened 2019-09-13
Improve vendoring script

aburgel:
/no-platform

1. Switch from bower to npm

2. Remove her/fakeable_her from script since the version that's pulled doesn't match what is already committed.

_update: removed the bit about adding the cli downloading, it was already there, i just bumped the version number_


samandmoore:
> 2\. Remove her/fakeable_her from script since the version that's pulled doesn't match what is already committed.

so you're saying that we patched these gems in this repo manually or something so you're just giving up on updating them for now? i'm fine with that

samandmoore:
<<domain tafn just wanna understand the gem thing

aburgel:
> so you're saying that we patched these gems in this repo manually or something so you're just giving up on updating them for now? i'm fine with that

yeah, basically. as is, you have to back out the changes made by this script, so it makes it not useful for updating.

there may be some commit hash that maps to the version we're actually using, but i'm not sure what that is. given that our usage of those libraries is pretty stable, i don't think there's any value in trying to update to the latest versions. our best bet would be to rip these out and use something more modern for rest apis.

aburgel:
bump

samandmoore:
domainlgtm

PR #93, opened 2019-09-19
Fix CLI failures when English module is missing

aburgel:
/no-platform
/cc @joejansen @fidelscodes

`TesttrackCli` module makes use of `$CHILD_STATUS`, which is provided by the `English` module. We weren't explicitly requiring it, so the code only happened to work because many gems `require 'English'` for their own purposes. The airbrake gem, which we carried as a dependency up until recently, is one example. If you're unlucky enough to not use such a gem, then `$CHILD_STATUS` will always be nil and blow up in our usecase.

I added a test that runs the CLI via `TesttrackCli`, hoping that it would fail, but it turns out that `simplecov` gem does a `require 'English'`, so `TesttrackCli#call` succeeds. I'm going to leave the test in anyways, since its a useful integration test.

### DIFF
--- a/app/models/testtrack_cli.rb
+++ b/app/models/testtrack_cli.rb
@@ -1,4 +1,5 @@
 require 'rbconfig'
+require 'English'
 
 class TesttrackCli
   include Singleton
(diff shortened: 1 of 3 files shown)


jmileham:
<< domain lgtm!

smudge:
What is `$?` not english enough?

PR #94, opened 2019-11-07
Introduce test_track_js_client 2.0 alpha with registry hash update

HipsterBrown:
/domain @aburgel @samandmoore @smudge 
/platform @aburgel @samandmoore @smudge 

Working off @aburgel's branch (https://github.com/Betterment/test_track_rails_client/tree/aburgel/js-2.0), this PR introduces the alpha version of `test_track_js_client` 2.0 for testing in our rails apps. Along with that updated library, the breaking contract changes split registry hash to match the Ruby implementation is also part of this PR to work with JS client 2.0.


Rakefile
```diff
@@ -28,12 +28,14 @@ RuboCop::RakeTask.new
 
 desc "Pull the latest versions of all dependencies into the gem for distribution"
 task :vendor_deps do
-  TEST_TRACK_JS_CLIENT_VERSION = '1.3.8'.freeze
+  TEST_TRACK_JS_CLIENT_VERSION = '2.0.0-alpha.2'.freeze
```
HipsterBrown: This will need to be updated once this (https://github.com/Betterment/test_track_js_client/pull/23) lands and we cut a `2.0.0-alpha.3` release.

Rakefile
```diff
@@ -28,12 +28,14 @@ RuboCop::RakeTask.new
...
   TEST_TRACK_CLI_VERSION = 'v1.0.2'.freeze
 
   # Bundle JS client
+  sh 'npm init -y'
```
HipsterBrown: Newer versions of `npm` require a `package.json` to be present when installed packages, so this temporarily creates one to be removed after the new client is copied into the vendored assets.

spec/models/test_track/web_session_spec.rb
```diff
@@ -513,17 +513,22 @@ def current_clown; end
     let(:v1_split_registry) do
```
HipsterBrown: Not sure to what we should rename this.

aburgel: we can just call it split registry now that there is only a single version

HipsterBrown:
I updated the vendored js_client to the latest release with the patch for AB Configuration and tested it locally in retail app with no errors. The Chrome extension has also been tested and is working as well. No issues with the new split registry format.

samandmoore:
<<domain

app/models/test_track/split_registry.rb
```diff
@@ -27,9 +27,9 @@ def weights_for(split_name)
     registry_hash && registry_hash['splits'][split_name] && registry_hash['splits'][split_name]['weights'].freeze
   end
 
-  def to_v1_hash
+  def to_hash
     registry_hash && registry_hash['splits'].each_with_object({}) do |(k, v), result|
-      result[k] = v['weights']
+      result[k] = { weights: v['weights'], feature_gate: v['feature_gate'] }
```
samandmoore: so the new JS client code consumes `splits` from `window.TT` in this new format?

HipsterBrown: Exactly. This is one of the breaking changes included in the 2.0.0 major update.

samandmoore:
domainlgtm!

HipsterBrown:
Bump

Still looking for platform review. 🙏

aburgel:
<< platform

spec/controllers/concerns/test_track/test_trackable_controller_spec.rb
```diff
@@ -11,7 +11,7 @@
 
     def index
       render json: {
-        v1_split_registry: test_track_session.state_hash[:registry],
+        v1_split_registry: test_track_session.state_hash[:splits],
```
aburgel: we should update the name of this key since its no longer v1

spec/models/test_track/web_session_spec.rb
```diff
@@ -539,18 +544,22 @@ def current_clown; end
...
 
     it "includes the assignment registry" do
       expect(subject.state_hash[:assignments]).to eq("assignments")
     end
 
+    it "includes the experience_sampling_weight" do
+      expect(subject.state_hash[:experienceSamplingWeight]).to eq(sampling_weight)
+    end
+
     it "includes a nil :registry if visitor returns a nil split_registry" do
```
aburgel:
```suggestion
    it "includes a nil :splits if visitor returns a nil split_registry" do
```

aburgel:
platformlgtm after those two small changes

HipsterBrown:
Changes have been implemented @aburgel

PR #95, opened 2020-01-03
Add special case for localhost cookie

chukkwagon:
/platform @aburgel @jmileham 
/cc @Betterment/security 

Adding a special case for localhost when setting the test track cookie domain. Currently, the cookie domain is set to `.localhost`, which is invalid in chrome and leads to all assignments reverting to the default. This is only a problem for developers using `rails s` instead of pow for local application development.

### DIFF
--- a/app/models/test_track/web_session.rb
+++ b/app/models/test_track/web_session.rb
@@ -94,7 +94,7 @@ class TestTrack::WebSession
   end
 
   def _cookie_domain
-    if bare_ip_address? || fully_qualified_cookie_domain_enabled?
+    if bare_ip_address? || request.local? || fully_qualified_cookie_domain_enabled?
       request.host
     else
       wildcard_domain
--- a/spec/models/test_track/web_session_spec.rb
+++ b/spec/models/test_track/web_session_spec.rb
@@ -19,8 +19,8 @@ RSpec.describe TestTrack::WebSession do
 
   let(:cookies) { { tt_visitor_id: "fake_visitor_id" }.with_indifferent_access }
   let(:headers) { {} }
-  let(:request) { double(:request, host: "www.foo.com", ssl?: true, headers: headers, local?: local_request?) }
-  let(:local_request?) { false }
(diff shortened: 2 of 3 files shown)


app/models/test_track/web_session.rb
```diff
@@ -93,7 +93,7 @@ def cookie_domain
   end
 
   def _cookie_domain
-    if bare_ip_address? || fully_qualified_cookie_domain_enabled?
+    if bare_ip_address? || request.local? || fully_qualified_cookie_domain_enabled?
```
jmileham: does this have weird interactions with consider_all_requests_local or anything?

jmileham: nope: https://github.com/rails/rails/blob/98a57aa5f610bc66af31af409c72173cdeeb3c9e/railties/lib/rails/application_controller.rb#L21-L23

spec/models/test_track/web_session_spec.rb
```diff
@@ -19,7 +19,8 @@ def current_clown; end
 
   let(:cookies) { { tt_visitor_id: "fake_visitor_id" }.with_indifferent_access }
   let(:headers) { {} }
-  let(:request) { double(:request, host: "www.foo.com", ssl?: true, headers: headers) }
+  let(:request) { double(:request, host: "www.foo.com", ssl?: true, headers: headers, local?: local_request?) }
+  let(:local_request?) { false }
```
jmileham: this naming is a little confusing though, given the `ApplicationController#local_request?` method which loops in the consider_all_requests_local config - maybe just call this `local?`

jmileham:
<< domain lgtm platform lgtm - minor naming tweak optional in test to avoid confusion with the more complicated behavior of `ApplicationController#local_request?`

spec/models/test_track/web_session_spec.rb
chukkwagon: You're right. I was trying to make it clear that this method was for the request only, but it introduces other ambiguity. Renaming it to `local?`.

PR #96, opened 2020-01-07
Add support for rails 6

RowanMcDonald:
/platform @aburgel @jmileham @effron @smudge

In the path to upgrading, I've been running around resolving a bit of dependency hell 🔥.

The diff here is a lot larger than it is significant.

# changes
1. update ruby-version to 2.6 which is required by rails 6 (well it requires 2.5+ but why not... 🤷‍♂)

1. ruby 2.6 is not supported by rubocop 0.51.0, so I updated that to 0.61.0

1. The newer version of rubocop has a lot of "suggestions" for this repo so I ran `bundle exec rubocop --auto-correct` and then `rm .rubocop_todo.yml; bundle exec rubocop --auto-gen-config` to ignore anything that wasn't auto corrected.

1. generate a rails 6 appraisal

1. update the build matrix to include the rails 6 appraisal.

# notes
1. If there's disagreement with the changes that rubocop --auto-correcting has introduced, I'm happy to shunt the entire thing into the todo.yml.

1. Also, I'm not very familiar with the internals of test track. If there are tests I should add so we can feel confident supporting rails 6, give me a shout!


.travis.yml
```diff
@@ -14,3 +15,11 @@ before_script:
 gemfile:
   - gemfiles/rails_5_1.gemfile
   - gemfiles/rails_5_2.gemfile
+  - gemfiles/rails_6_0.gemfile
+
+matrix:
+  allow_failures:
```
smudge: This can be `exclude:`

smudge:
Changing the min required ruby version is probably a breaking change, right? We can probably do away with that for the sake of existing apps on older rubies, right?

<< platform TAFN

RowanMcDonald:
bump, I reverted updating the min required ruby and updated the travis config.

smudge:
<< domain LGTM && platform LGTM

PR #97, opened 2020-01-31
Add support for timestamped split registry endpoint

reggieag:
/platform @aburgel @jmileham @effron @smudge

This PR adds the build timestamp to the rails client. We decided to push the timestamp generation and validation up to the build of the client so the app code can assume that the timestamp exists.


app/models/test_track/remote/split_registry.rb
```diff
@@ -15,7 +15,7 @@ def instance
       if faked?
         new(fake_instance_attributes(nil))
       else
-        get('/api/v2/split_registry')
+        get("/api/v3/builds/#{TestTrack::BuildTimestamp}/split_registry")
```
aburgel: We may be able to pass in the timestamp as an argument to `get`. Though we don’t need url encoding since the timestamp shouldn’t have any funky characters, I think it’s a good practice.

app/models/test_track/remote/split_registry.rb
```diff
@@ -3,7 +3,7 @@ class TestTrack::Remote::SplitRegistry
 
   CACHE_KEY = 'test_track_split_registry'.freeze
 
-  collection_path '/api/v2/split_registry'
+  collection_path '/api/v3/builds/:build_timestamp/split_registry'
```
aburgel: Is this used in any way? I think every usecase should go through the instance method below but we should confirm.

config/initializers/test_track_generate_timestamp.rb
```diff
@@ -0,0 +1,9 @@
+BUILD_TIMESTAMP_FILE_PATH = 'testtrack/build_timestamp'.freeze
+
+if Rails.env.test? || Rails.env.development?
+  TestTrack::BuildTimestamp = Time.zone.now.iso8601
+elsif File.exist?(BUILD_TIMESTAMP_FILE_PATH) && File.readable?(BUILD_TIMESTAMP_FILE_PATH) && !File.zero?(BUILD_TIMESTAMP_FILE_PATH)
+  TestTrack::BuildTimestamp = File.read(BUILD_TIMESTAMP_FILE_PATH)
+else
+  raise Exception.new "Failed to read timestamp"
```
aburgel: This could use more context. If the app fails to startup we should provide plenty of details on why and how to fix.

Something like: “TestTrack failed to load the required build timestamp. Ensure testtrack-cli build-timestamp is run and the timestamp file is present”

Something like that but better 😁

config/initializers/test_track_generate_timestamp.rb
```diff
@@ -0,0 +1,9 @@
+BUILD_TIMESTAMP_FILE_PATH = 'testtrack/build_timestamp'.freeze
+
+if Rails.env.test? || Rails.env.development?
+  TestTrack::BuildTimestamp = Time.zone.now.iso8601
+elsif File.exist?(BUILD_TIMESTAMP_FILE_PATH) && File.readable?(BUILD_TIMESTAMP_FILE_PATH) && !File.zero?(BUILD_TIMESTAMP_FILE_PATH)
+  TestTrack::BuildTimestamp = File.read(BUILD_TIMESTAMP_FILE_PATH)
```
aburgel: I think some basic validation of the file content is worthwhile, like is it in iso format. @jmileham does the iOS client do any validation?

lib/tasks/test_track_rails_client_tasks.rake
```diff
@@ -10,6 +10,17 @@ namespace :test_track do
     end
   end
 
+  desc 'Generates build timestamp'
+  task generate_build_timestamp: :environment do
+    cli = TesttrackCli.instance
+    next if cli.skip_testtrack_cli?
+
+    if cli.project_initialized?
+      result = cli.call('generate_timestamp')
```
aburgel: Hmm, the naming of this command and the rake task is inconsistent. I like the more explicit name of the rake task, so probably should change the name of this command to `generate_build_timestamp`.

lib/tasks/test_track_rails_client_tasks.rake
```diff
@@ -31,3 +42,4 @@ end
 task 'db:schema:load' => ['test_track:schema:load']
 task 'db:structure:load' => ['test_track:schema:load']
 task 'db:migrate' => ['test_track:migrate']
+task 'deploy:assets:precompile' => ['test_track:generate_build_timestamp']
```
aburgel: Would be nice to add a test for this since it’s now pretty critical to building the apps, not sure if we have any examples of testing rake tasks.

config/initializers/test_track_generate_timestamp.rb
```diff
@@ -0,0 +1,9 @@
+BUILD_TIMESTAMP_FILE_PATH = 'testtrack/build_timestamp'.freeze
+
+if Rails.env.test? || Rails.env.development?
```
aburgel: Getting some tests around this behavior would be great too. Not sure if you can test an initializer. But we could pull this logic into a function, test that and then just call that here

aburgel:
<< platform

aburgel:
Tafn! This is great. Mostly comments about test coverage. Nice work!

app/models/test_track/remote/split_registry.rb
phantomphildius: both https://github.com/reggieag/test_track_rails_client/blob/rg/add-timestamp/app/models/test_track/remote/visitor_detail.rb and https://github.com/reggieag/test_track_rails_client/blob/rg/add-timestamp/app/models/test_track/remote/visitor.rb follow this pattern but the rest of the remote models rely only on collection_path. I wonder when we wanted to start potentially ignoring collection_path. is it some sort of HER framework requirement or a documentation like attribute? It does seem superfluous if we're just using the call on line 18

lib/tasks/test_track_rails_client_tasks.rake
phantomphildius: so update the command in the cli?

app/models/test_track/remote/split_registry.rb
jmileham: Her doesn't actually support singular routes I think, so we duplicate when we have to IIRC, just so it's clear what the route is we're talking about.

config/initializers/test_track_generate_timestamp.rb
jmileham: I'm not sure, but agreed. Also, this camelization isn't conventional for ruby - it looks like a class. We should either have a property of `TestTrack`, a la `TestTrack.build_timestamp`, or it should be using constant case, i.e. `TestTrack::BUILD_TIMESTAMP`. If we don't intend for it to be mutable, we should do that. Also, the timestamp file should probably be newline terminated, and if so, this should probably be chomping the results during its validation/normalization phase.

config/initializers/test_track_generate_timestamp.rb
jmileham: Any reason we're raising a bare exception instead of just the text, which would result in a StandardError?

app/models/test_track/remote/split_registry.rb
phantomphildius: Seems like we're doing string interpolation in the other instances of get with args. Should we stick with that pattern or forge our own path?

phantomphildius:
nanda

config/initializers/test_track_load_timestamp.rb
```diff
@@ -0,0 +1 @@
+TestTrack.load_build_timestamp
```
aburgel: do we need `test_track` in the filename? may just `load_build_timestamp`? since test_track is kinda implied

lib/test_track.rb
```diff
@@ -52,6 +54,23 @@ def misconfiguration_notifier_class_name=(notifier_class_name)
...
+    def load_build_timestamp
+      if Rails.env.test? || Rails.env.development?
+        @build_timestamp = Time.zone.now.iso8601
+      elsif File.exist?(BUILD_TIMESTAMP_FILE_PATH)
+        raise "TestTrack's build_timestamp is not formatted properly." unless BUILD_TIMESTAMP_REGEX.match?(_build_timestamp)
+
+        @build_timestamp = _build_timestamp
+      else
+        raise 'TestTrack failed to load the required build timestamp. ' \
+          'Ensure `testtrack load_build_timestamp` is run and the build timestamp file is present.'
```
aburgel:
```suggestion
          'Ensure `test_track:generate_build_timestamp` task is run and the build timestamp file is present.'
```

lib/test_track.rb
```diff
@@ -52,6 +54,23 @@ def misconfiguration_notifier_class_name=(notifier_class_name)
       @misconfiguration_notifier_class_name = notifier_class_name
     end
 
+    def load_build_timestamp
+      if Rails.env.test? || Rails.env.development?
+        @build_timestamp = Time.zone.now.iso8601
+      elsif File.exist?(BUILD_TIMESTAMP_FILE_PATH)
+        raise "TestTrack's build_timestamp is not formatted properly." unless BUILD_TIMESTAMP_REGEX.match?(_build_timestamp)
```
aburgel:
```suggestion
        raise "./testtrack/build_timestamp is not a valid ISO 8601 timestamp, got '#{_build_timestamp}'" unless BUILD_TIMESTAMP_REGEX.match?(_build_timestamp)
```

lib/test_track.rb
```diff
@@ -52,6 +54,23 @@ def misconfiguration_notifier_class_name=(notifier_class_name)
...
+
+        @build_timestamp = _build_timestamp
+      else
+        raise 'TestTrack failed to load the required build timestamp. ' \
+          'Ensure `testtrack load_build_timestamp` is run and the build timestamp file is present.'
+      end
+    end
+
+    def build_timestamp
+      @build_timestamp.presence || raise('build_timestamp is not defined. Ensure `load_build_timestamp` is run.')
```
aburgel:
```suggestion
      @build_timestamp.presence || raise('build_timestamp is not defined because `load_build_timestamp` initializer was not run.')
```

aburgel:
platformlgtm! looking great! dropped in a couple of minor suggestions around the exception messages.

phantomphildius:
nanda

lib/test_track.rb
```diff
@@ -52,6 +54,27 @@ def misconfiguration_notifier_class_name=(notifier_class_name)
...
+
+        @build_timestamp = timestamp
+      else
+        raise 'TestTrack failed to load the required build timestamp. ' \
+          'Ensure `test_track:generate_build_timestamp` task is run and the build timestamp file is present.'
+      end
+    end
+
+    def build_timestamp
+      @build_timestamp.presence || raise('build_timestamp is not defined because `load_build_timestamp` initializer was not run.')
```
smudge: I wonder if this should note that it normally runs as part of `assets:precompile`

smudge: (same as the error above)

lib/test_track.rb
```diff
@@ -52,6 +54,27 @@ def misconfiguration_notifier_class_name=(notifier_class_name)
       @misconfiguration_notifier_class_name = notifier_class_name
     end
 
+    def load_build_timestamp # rubocop:disable Metrics/MethodLength
+      if Rails.env.test? || Rails.env.development?
+        @build_timestamp = Time.zone.now.iso8601
+      elsif File.exist?(BUILD_TIMESTAMP_FILE_PATH)
```
smudge: I'm thinking we can maybe combine the error cases so we only have to `raise` once (regardless of whether the file exists)


```ruby
    def load_build_timestamp
      if Rails.env.test? || Rails.env.development?
        @build_timestamp = Time.zone.now.iso8601
      elsif _build_timestamp
        unless BUILD_TIMESTAMP_REGEX.match?(timestamp)
          raise "./testtrack/build_timestamp is not a valid ISO 8601 timestamp, got '#{timestamp}'"
        end

        @build_timestamp = _build_timestamp
      else
        raise '...'
      end
    end

    attr_reader :build_timestamp

  # ...

  def _build_timestamp
    File.read(BUILD_TIMESTAMP_FILE_PATH).chomp.presence if File.exist?(BUILD_TIMESTAMP_FILE_PATH)
  end
```

Main thing there is that by calling `.presence` sooner and sharing the `raise` during the initializer code, we can make `build_timestamp` an `attr_reader` and read straight off the ivar.

smudge: (I'm thinking this is a safe strategy because the gem itself defines the initializer and owns `load_build_timestamp`, so unless we suspect that somehow folks are gonna skip initialization of the app?)

smudge: (Alternatively, I sort of wonder if we could rename `load_build_timestamp` to `build_timestamp` -- have it fetch all the things internally, run all the checks, etc -- and then basically just pre-heat it from the initializer.)

smudge:
<< domain LGTM - this looks totally correct to me. I don't want to block you from merging, but I did leave a pile o' structural suggestions, take them or leave them. (Also I'm thinking we could mention the `assets:precompile` behavior in the error messages.)

lib/tasks/test_track_rails_client_tasks.rake
```diff
@@ -31,3 +39,4 @@ end
 task 'db:schema:load' => ['test_track:schema:load']
 task 'db:structure:load' => ['test_track:schema:load']
 task 'db:migrate' => ['test_track:migrate']
+task 'assets:precompile' => ['test_track:generate_build_timestamp']
```
phantomphildius: going to try reordering the tasks and adding a flag to not run the initializers until this is done here

phantomphildius:
nanda

PR #98, opened 2020-04-14
Fast forward Ruby support to 2.5-2.7

smudge:

This ups the required Ruby version in the gemspec from 2.1.0 to 2.5.0 -- In practice I don't know how we were supporting Ruby 2.1.0 if our minimum tested Rails version was 5.1, and given that 2.4 and below have been EOL'd, it feels safe to say we only support [maintained ruby versions](https://www.ruby-lang.org/en/downloads/branches/).


aburgel:
<< domainlgtm platformlgtm

aburgel:
Just realized that you asked for two reviews, probably wanting to get consensus on this change. My bad for grabbing both! @samandmoore @jmileham please weigh in!

jmileham:
<< domain lgtm too!

PR #99, opened 2020-04-15
Rework/simplify development dependencies

smudge:
/no-platform

Seems like a bunch of development dependencies had accrued in the `Gemfile` over time -- after some experimentation, it seems like only one was actually necessary (`jbuilder`) so I added a comment explaining why, and put all of the others in the `.gemspec`. (Actually, `selenium-webdriver` seemed entirely unused, so I ended up removing it.)

This results in tidier Appraisal-generated gemfiles too.

Again, development dependencies -- this shouldn't impact the actual release cut of the gem.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,18 +1,13 @@
 appraise 'rails-5-1' do
   gem 'rails', '~> 5.1.0'
-  gem 'rack'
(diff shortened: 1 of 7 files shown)


aburgel:
<< domainlgtm

PR #100, opened 2020-04-15
Allow TT API to live on non-root path

smudge:
/platform @Betterment/test_track_core @effron 

**Note: this depends on #98 so I've rebased on that branch for the time being.**

### The actual change: https://github.com/Betterment/test_track_rails_client/commit/2002c14e5d594f1a95bc6938c5c951ec32be067d

TLDR; if your TEST_TRACK_API_URL is not at the root of the host -- e.g. if it's something like `https://something.test/tt` -- the `/tt` will get dropped by the client because it was doing:

```ruby
get("/api/path/to/stuff")

# resulting in https://something.test/api/path/to/stuff
```

I am changing it to:

```ruby
get("api/path/to/stuff")

# resulting in https://something.test/tt/api/path/to/stuff
```

(notice the lack of a leading `/`)


aburgel:
<< domainlgtm looks reasonable to me!

effron:
<<platformLGTM!

PR #101, opened 2020-04-15
Fix Ruby 2.7 build, clear out .rubocop_todo.yml

smudge:

In order to support Ruby 2.7 syntax, I needed to upgrade rubocop.

While I was upgrading rubocop, I went ahead and auto-corrected, which left me with a handful of failures to manually correct. I'll annotate the interesting ones below.


.rubocop.yml
```diff
@@ -1,8 +1,9 @@
 # See full list of defaults here: https://github.com/bbatsov/rubocop/blob/master/config/default.yml
 # To see all cops used see here: https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml
 
-inherit_from:
-  - .rubocop_todo.yml
+require:
+  - rubocop-rails
+  - rubocop-performance
```
smudge, line 6: These were both extracted into separate gems but had previously been running in this repo.

.rubocop.yml
```diff
@@ -24,9 +25,12 @@ Layout/CaseIndentation:
 Layout/ClosingParenthesisIndentation:
   Enabled: false
 
-Layout/IndentArray:
+Layout/FirstArrayElementIndentation:
   EnforcedStyle: consistent
 
+Layout/LineLength:
+  Max: 140
```
smudge, line 32: Moved from below (from "Metrics" to "Layout")

.rubocop.yml
```diff
@@ -92,12 +99,20 @@ Rails/Delegate:
 Rails/FindEach:
   Enabled: false
 
+Rails/HasManyOrHasOneDependent:
+  Exclude:
+    - 'app/models/test_track/remote/**/*'
```
smudge, line 104: Excluding the Her models from this cop. Maybe we can disable it entirely 🤷

app/models/test_track/ab_configuration.rb
```diff
@@ -20,7 +20,7 @@ def variants
 
   def build_variant_hash
     notify_because_ab("configures split with more than 2 variants") if split_variants && split_variants.size > 2
-    { true: true_variant, false: false_variant }
+    { true: true_variant, false: false_variant } # rubocop:disable Lint/BooleanSymbol
```
smudge, line 23: `disable Lint/BooleanSymbol` appears all over the place -- we can just disable it globally, but then we lose the cop in places where we might actually mean to pass a boolean value... 😢 

If only we could add a little bit of type-casting to variant values, like, just as syntactic sugar for all the consuming code... 🤔

app/models/test_track/config_updater.rb
```diff
@@ -1,5 +1,5 @@
 class TestTrack::ConfigUpdater
-  def initialize(schema_file_path = Rails.root.join('db', 'test_track_schema.yml'))
+  def initialize(schema_file_path = Rails.root.join('db/test_track_schema.yml'))
```
smudge, line 2: I feel like we use the reverse of this config elsewhere -- I intentionally avoided pulling in `rubocop-betterment` to this repo to avoid blowing up too many things at once, but maybe it's time?

app/models/test_track/lazy_visitor_by_identity.rb
```diff
@@ -14,15 +14,19 @@ def id_loaded?
   private
 
   def method_missing(method, *args, &block)
-    __visitor__.send(method, *args, &block)
+    if __visitor__.respond_to?(method)
+      __visitor__.send(method, *args, &block)
+    else
+      super
```
smudge, line 20: Rubocop wants you to fall back on `super` from method_missing. I get why it's valuable in most cases. In this case, it means that we get a slightly different stacktrace when a method truly isn't defined on the visitor class. 🤔

app/models/test_track/split_registry.rb
```diff
@@ -28,8 +28,8 @@ def weights_for(split_name)
   end
 
   def to_hash
-    registry_hash && registry_hash['splits'].each_with_object({}) do |(k, v), result|
-      result[k] = { weights: v['weights'], feature_gate: v['feature_gate'] }
+    registry_hash && registry_hash['splits'].transform_values do |v|
+      { weights: v['weights'], feature_gate: v['feature_gate'] }
```
smudge, line 32: ✨

app/models/test_track/visitor.rb
```diff
@@ -39,18 +39,18 @@ def ab(split_name, opts = {}) # rubocop:disable Metrics/MethodLength, Metrics/Ab
...
         false
       end
     end
   end
 
   def assignment_registry
-    @assignment_registry ||= assignments.each_with_object({}) do |assignment, hsh|
-      hsh[assignment.split_name] = assignment
+    @assignment_registry ||= assignments.index_by do |assignment|
+      assignment.split_name
```
smudge, line 53: ✨

lib/tasks/test_track_rails_client_tasks.rake
```diff
@@ -19,7 +19,7 @@ namespace :test_track do
   end
 
   desc 'Sets an environment variable to block build timestamp generation on application initialization'
-  task :skip_load_build_timestamp do
+  task :skip_load_build_timestamp do # rubocop:disable Rails/RakeEnvironment
     ENV['SKIP_TESTTRACK_SET_BUILD_TIMESTAMP'] = '1'
```
smudge, line 23: I wasn't sure if we intentionally didn't inherit from `:environment` here -- cc @phantomphildius

spec/models/test_track/remote/split_registry_spec.rb
```diff
@@ -22,7 +22,7 @@
 
     begin
       Rails.cache.clear
-    rescue Errno::ENOENT # rubocop:disable Lint/HandleExceptions
+    rescue Errno::ENOENT
       # This is fine
```
smudge, line 26: You're now allowed to comment about _why_ you aren't handling an exception, and it doesn't violate the cop.

spec/rails_helper.rb
```diff
@@ -14,7 +14,7 @@
 require 'timecop'
 
 # load all files in support folders
-Dir[TestTrackRailsClient::Engine.root.join("spec/support/**/*.rb")].each { |f| require f }
+Dir[TestTrackRailsClient::Engine.root.join("spec/support/**/*.rb")].sort.each { |f| require f }
```
smudge, line 17: `Dir`'s default sorting is OS/filesystem-dependent, so this is a pretty smart rule.

spec/rails_helper.rb
```diff
@@ -60,6 +60,7 @@
   # https://relishapp.com/rspec/rspec-rails/docs
   config.infer_spec_type_from_file_location!
 
+  config.include Shoulda::Matchers::Independent
```
smudge, line 63: The newer shoulda-matchers versions inject specific matchers for specific types of tests, but it only works from within the context of a real Rails app apparently 🤷

lib/tasks/test_track_rails_client_tasks.rake
phantomphildius: This was when we were trying to avoid loading all the other tasks that would have gotten called via environment. We can probably add it back now!

jmileham:
<< domain lgtm platform lgtm

lib/tasks/test_track_rails_client_tasks.rake
smudge: Cool, will add it back.

phantomphildius: does this seem right @aburgel ?

PR #102, opened 2020-04-16
Bump version to 4.0.0.rc1

aburgel:
/no-platform
/cc @phantomphildius @reggieag 

Also incorporated some upgrade notes for the new build timestamp feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Set up ENV vars in every environment:
 * `TEST_TRACK_API_URL` - Set this to the URL of your TestTrack instance with your app credentials, e.g. `http://[myapp]:[your new app password]@[your-app-domain]/`
 
   [your-app-domain] can be
-  * `testtrack.dev` ([Pow](pow.cx))
+  * `testtrack.test`
   * `localhost:PORT`
   * `example.org`
   * etc
@@ -72,8 +72,8 @@ If your app doesn't support authentication, set
 
 ### Prepare your identity models (optional)
 
-If your app supports authentication, You'll need to configure your
-`User` model as a [TestTrack Identity](#varying-app-behavior-from-within-a-model)
+If your app supports authentication, you'll need to configure your
+`User` model as a [TestTrack Identity](#varying-app-behavior-from-within-a-model).
 
 ### Set up the Chrome extension (optional)
 
@@ -288,7 +288,7 @@ class BackgroundWorkJob
     end
   end
 end
-``` 
+```
 
 ## Tracking visitor logins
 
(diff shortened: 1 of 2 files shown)


lib/test_track_rails_client/version.rb
```diff
@@ -1,3 +1,3 @@
 module TestTrackRailsClient
-  VERSION = "4.0.0.alpha35" # rubocop:disable Style/MutableConstant
+  VERSION = "4.0.0-rc1" # rubocop:disable Style/MutableConstant
```
aburgel, line 2: I think we weren't semver compliant by using a `.` instead of `-` 🤷

README.md
```diff
@@ -391,28 +391,29 @@ def notify(message)
...
+* You must add `self.test_track_identity = :current_user` (or
 whatever your controller uses as a sign-in identity) to your
 TestTrack-enabled controllers, or set it to `:none` if your app doesn't
-support authentication.
+support authentication. If your app supports authentication, you'll need to configure your
+user model as a [TestTrack Identity](#varying-app-behavior-from-within-a-model).
 
-If your app supports authentication, You'll need to configure your
-user model as a [TestTrack Identity](#varying-app-behavior-from-within-a-model)
+* TestTrack server [introduced a new endpoint](https://github.com/Betterment/test_track/pull/133) for fetching the split registry that requires a timestamp of when the application was built. This prevents dropped splits from breaking a running application. If you use `rake assets:precompile` in your build pipeline, you're all set. If you don't, you'll need to call `rake test_track:generate_build_timestamp` and ensure that the file `testtrack/build_timestamp` is deployed along with your app.
```
aburgel, line 406: All except this line are minor formatting & typo changes.

lib/test_track_rails_client/version.rb
jmileham: i think there's something magic about the dot for gems - be careful that this doesn't cut a non-prerelease.

aburgel: oh interesting! i assumed that semver rules were the standard, i'll do a little more research.

jmileham: Rails does RC's with dots, for example: https://rubygems.org/gems/rails

lib/test_track_rails_client/version.rb
```diff
@@ -1,3 +1,3 @@
 module TestTrackRailsClient
-  VERSION = "4.0.0.alpha35" # rubocop:disable Style/MutableConstant
+  VERSION = "4.0.0-rc.1" # rubocop:disable Style/MutableConstant
```
smudge, line 2: Based on other gem conventions I think this should be `4.0.0.rc1`

smudge:
<< domain TAFN - just the one thing. Exciting!!! 😄

lib/test_track_rails_client/version.rb
aburgel: and here i thought i was making the world a better place. oh well, back to dots!

aburgel:
bump!

lib/test_track_rails_client/version.rb
smudge: Apparently [all you need are one or more letters in the gem version](https://guides.rubygems.org/patterns/#prerelease-gems), but all the examples they include split it with a "." after the minor version, and it seems like a somewhat nuanced parsing problem. So, I figure if Rails does it this way, it's a pretty safe bet.

smudge:
domain LGTM! :shipit: 🚀

lib/test_track_rails_client/version.rb
jmileham: I think it might be for collation order - e.g. you want `4.0.1.rc1` to sort just before `4.0.1`, not before `4.0`, which it might do given the parsing logic in gems.

PR #103, opened 2020-04-17
Fix 'rake assets:precompile'

smudge:
/domain @aburgel @phantomphildius 
/no-platform

This diff looks crazier than it is. (Try it with `?w=1`)

Basically:

**Step 1:** Get actual `rake assets:precompile` command working in this repo via the rspec "dummy" app. 👍 

_We need to run the command in a production ENV to fully exercise the non-local behavior, which sounds scary, but remember this is an empty gem, so I made `production.rb` a copy of `test.rb`_

**Step 2:** Write a test that attempts to fully execute the rake precompile command and checks that it sets the build timestamp (and returns a successful exit code). 👍 

_This also feels like a lot since it calls out via `system`, but think of this as an integration test -- it's a surefire way to ensure that all of the initializers fully execute, which they won't do if we call the task directly via `Rake::Task[].invoke`!_

**Step 2.5:** Confirm that the test _FAILS_ 🔴 

**Step 3:** Undo the thing rubocop suggested in the 2.7 fix PR, and confirm that the test from step 2 now passes ✅ 

_Also make sure we cleanup the very real `testtrack/` dir that the thing creates._

...

🎉

### DIFF
--- a/spec/test_track_spec.rb
+++ b/spec/test_track_spec.rb
@@ -276,23 +276,25 @@ RSpec.describe TestTrack do
           end
         end
       end
+    end
 
-      context 'when set via assets:precompile' do
-        after do
-          Rake::Task['app:assets:clobber'].invoke
-          FileUtils.remove_dir('testtrack', true)
-        end
+    context 'when set via full assets:precompile command' do
+      around do |example|
+        FileUtils.remove_dir('testtrack', true)
+        example.run
+        Rake::Task['app:assets:clobber'].invoke
+        FileUtils.remove_dir('testtrack', true)
+      end
 
-        let(:asset_precompile_success) do
-          system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake app:assets:precompile')
-        end
+      let(:asset_precompile_success) do
+        system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake app:assets:precompile')
+      end
 
-        it 'does not raise an error' do
-          expect { asset_precompile_success }
-            .to change { File.exist?('testtrack/build_timestamp') }
-            .from(false).to(true)
-          expect(asset_precompile_success).to eq true
-        end
+      it 'does not raise an error' do
+        expect { asset_precompile_success }
+          .to change { File.exist?('testtrack/build_timestamp') }
+          .from(false).to(true)
+        expect(asset_precompile_success).to eq true
       end
     end
   end
--- a/spec/test_track_spec.rb
+++ b/spec/test_track_spec.rb
@@ -289,7 +289,7 @@ RSpec.describe TestTrack do
 
         it 'does not raise an error' do
           expect { asset_precompile_success }
-            .to change { File.exists?('testtrack/build_timestamp') }
+            .to change { File.exist?('testtrack/build_timestamp') }
             .from(false).to(true)
           expect(asset_precompile_success).to eq true
         end
--- a/spec/test_track_spec.rb
+++ b/spec/test_track_spec.rb
@@ -280,14 +280,16 @@ RSpec.describe TestTrack do
 
     context 'when set via full assets:precompile command' do
       around do |example|
-        FileUtils.remove_dir('testtrack', true)
-        example.run
-        Rake::Task['app:assets:clobber'].invoke
-        FileUtils.remove_dir('testtrack', true)
+        Dir.chdir('spec/dummy') do
+          File.delete('testtrack/build_timestamp') if File.exist?('testtrack/build_timestamp')
+          example.run
+          system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake assets:clobber')
+          File.delete('testtrack/build_timestamp') if File.exist?('testtrack/build_timestamp')
+        end
       end
 
       let(:asset_precompile_success) do
-        system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake app:assets:precompile')
+        system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake assets:precompile')
       end
 
       it 'does not raise an error' do
--- a/spec/test_track_spec.rb
+++ b/spec/test_track_spec.rb
@@ -203,7 +203,9 @@ RSpec.describe TestTrack do
       let(:file_readable) { true }
 
       before do
-        allow(File).to receive(:exist?).and_return(file_readable)
+        allow(File).to receive(:exist?).and_call_original
+        allow(File).to receive(:exist?).with('testtrack/build_timestamp').and_return(file_readable)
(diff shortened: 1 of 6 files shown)


spec/test_track_spec.rb
```diff
@@ -275,5 +277,25 @@
...
+        system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake app:assets:precompile')
+      end
+
+      it 'does not raise an error' do
+        expect { asset_precompile_success }
+          .to change { File.exist?('testtrack/build_timestamp') }
+          .from(false).to(true)
+        expect(asset_precompile_success).to eq true
+      end
+    end
```
aburgel, line 299: I think there's a more surgical way of doing this. If you `cd` into the dummy app, then you can avoid mucking with the Rakefile and all the other rake loading things you had to do. It also means that you are operating with that app's `testtrack` directory, rather than a fake `testtrack` directory at the root of the project. With just this change plus turning off uglifier if the dummy app's prod config, I was able to get all the tests passing, and it feels closer to a real world scenario.

```suggestion
    context 'when set via full assets:precompile command' do
      around do |example|
        Dir.chdir('spec/dummy') do
          File.delete('testtrack/build_timestamp') if File.exist?('testtrack/build_timestamp')
          example.run
          system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake assets:clobber')
          File.delete('testtrack/build_timestamp') if File.exist?('testtrack/build_timestamp')
        end
      end

      let(:asset_precompile_success) do
        system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake assets:precompile')
      end

      it 'does not raise an error' do
        expect { asset_precompile_success }
          .to change { File.exist?('testtrack/build_timestamp') }
          .from(false).to(true)
        expect(asset_precompile_success).to eq true
      end
    end
```

spec/test_track_spec.rb
```diff
@@ -275,5 +277,25 @@
         end
       end
     end
+
+    context 'when set via full assets:precompile command' do
+      around do |example|
+        FileUtils.remove_dir('testtrack', true)
+        example.run
+        Rake::Task['app:assets:clobber'].invoke
```
aburgel, line 285: Should we hook into `assets:clobber` to remove the `testtrack/build_timestamp` file? Feels like that would match user expectations. (Maybe in a later PR so we don't introduce too much entropy here.)

aburgel:
<< domain tafn!

This is awesome, @phantomphildius and I talked about how we could test this but shied away since it seemed gnarly.

jmileham:
Random q - does the timestamp get gitignored automatically, or are we just avoiding writing it in nonprod environments so devs don’t accidentally commit it? Does assets:clean clean it if we’re burning it out via assets:precompile?

jmileham:
(I think gitignoring would be more satisfying than only writing in production if we’re not doing that already)

jmileham:
Also should we yank rc1 if it’s totally broke?

aburgel:
> does the timestamp get gitignored automatically, or are we just avoiding writing it in nonprod environments so devs don’t accidentally commit it?

the latter... i was planning to add it to gitignore when i upgrade the apps. is there a way to automatically gitignore? that sounds much better but i don't know how that's done

> Does assets:clean clean it if we’re burning it out via assets:precompile?

not yet... i suggested it [above](https://github.com/Betterment/test_track_rails_client/pull/103#discussion_r410119884). i think we should... tho i think it should be `assets:clobber`

> should we yank rc1 if it’s totally broke?

good idea. doing that now.

aburgel:
yanked!

```sh
➜ gem yank test_track_rails_client --version 4.0.0.rc1
Yanking gem from https://rubygems.org...
Successfully deleted gem: test_track_rails_client (4.0.0.rc1)
```

jmileham:
You can put gitignore files anywhere in the tree so it probably would make the most sense for TestTrack CLI to write a gitignore ignoring the timestamp file whenever it is checking for the existence of a TestTrack dir, which will usually be when you write your first migration I think? In the meantime we can put similar behavior in the rails client in the timestamp cutting rake task?

spec/test_track_spec.rb
smudge: Okay! Happy to do that if it feels less invasive!

I've been taking a strategy with some gems where I enable full functionality at the root, including a `rackup` that lets me actually boot the dummy app and click around a test server. But I can understand how that would not be the goal of this PR. 🙂

spec/test_track_spec.rb
smudge: Yup, and we should **not** hook into `assets:clean` since it's common to run that directly after precompile to clear out older builds.

spec/test_track_spec.rb
smudge: And to clarify, would you prefer I leave the prod config intact and _just_ turn off uglifier?

aburgel:
> it probably would make the most sense for TestTrack CLI to write a gitignore ignoring the timestamp file whenever it is checking for the existence of a TestTrack dir 

I added it to the `init_project` command: https://github.com/Betterment/testtrack-cli/pull/55

That feels less surprising than having a .gitignore file appear when you run a migration. For existing projects, I think its fine to provide guidance to ignore the file. (I can add that to the readme when we bump the embedded cli version)

spec/test_track_spec.rb
```diff
@@ -275,5 +277,27 @@
         end
       end
     end
+
+    context 'when set via full assets:precompile command' do
+      around do |example|
+        Dir.chdir('spec/dummy') do
+          File.delete('testtrack/build_timestamp') if File.exist?('testtrack/build_timestamp')
+          example.run
+          system({ 'RAILS_ENV' => 'production' }, 'bundle exec rake assets:clobber')
```
smudge, line 286: @aburgel technically this is more symmetric and ensures that any production-related assets get clobbered, but in practice it's slower than `rm`ing the `dummy/public/assets` dir. (This is why I had relied on `Rake::Task` in the prior implementation.)

But unless we feel like the test time is particularly painful for this repo (which it isn't) my vote would be to run the full rake command. Thoughts?

smudge:
bump - assuming tests pass I think this has addressed your feedback @aburgel

spec/test_track_spec.rb
aburgel: I agree, especially if we plan to hook into `assets:clobber` to remove the file. We can just kill the following line.

Even better, we can move the clobber call into the test body and assert that the file is removed. Or is that overloading that test too much?

aburgel:
domainlgtm! this turned out really nice and surfaced some good developer experience improvements.

i can update the cli and add the `assets:clobber` hook in another PR and then we can cut rc2.

PR #108, opened 2020-12-03
Update js client to 2.0.0-alpha.6 and cli to 1.1.3

aburgel:
/no-platform 

The js update includes a bug fix as well as moves the underlying code to typescript. The cli update includes support for a new batch update endpoint.


smudge:
<< domain LGTM! Just a note that it's worth spot-checking retail.betterment.io and making sure that the JS client and chrome extension work with the self-hosted test track mounted at `/tt` (i.e. the `FakeTestTrack` webvalve app)

smudge:
<< domain LGTM!

PR #109, opened 2021-01-27
Unpin faraday dependency, support >= 1.0

smudge:
### Summary

I think this gem predates faraday 1.0 and we were simply being cautious, but now that 1.0 is out, it looks like we can unpin the version and all tests still pass.

### Other Information

I can confirm that faraday v1.3.0 is picked up by the lockfiles when bundle resolves dependencies. (I don't think there's a need to add an appraisal to test with an older faraday version, or anything like that.)

/domain @Betterment/test_track_core @RowanMcDonald 
/platform @Betterment/test_track_core @RowanMcDonald

### DIFF
--- a/test_track_rails_client.gemspec
+++ b/test_track_rails_client.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'delayed_job', '~> 4.0'
   s.add_dependency 'delayed_job_active_record'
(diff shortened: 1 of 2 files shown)


RowanMcDonald:
<< platformLGTM!

aburgel:
<< domainlgtm

nanda-prbot (approved):
Approved! :sparkler: :guitar: :shipit:

smudge:
Merging without a version bump!

PR #111, opened 2021-02-01
Add support for rails 6.1 and zeitwerk.

RowanMcDonald:
### Summary

This pr adds support for zeitwerk and adds an appraisal for rails 6.1. Also it adds `load_defaults` to the dummy application.rb.

To support zeitwerk, rails new autoloader, we need to make sure that rails can infer the name of the constant defined in each file by camelizing the file name. This is roughly the reverse of the previous behavior although the difference is sometimes subtle.

I think we have three options here:
1. Add an inflection to camelize dsl -> DSL.  I think the only way to do this is global and I don't think this gem should decide the global inflection.
2. Change the constant (this is what I have done.)
3. Convert this project from an engine to a railtie and circumvent zeitwerk all together (or use it explicitly to load things in isolation)

I have chosen 2 because it seems the most lightweight change.  Constants changed are:
1. `VaryDSL -> VaryDsl`
2. `VisitorDSL -> VisitorDsl`
3. `ABConfiguration -> AbConfiguration`
What do ya'll think?

/domain @Betterment/test_track_core @smudge 


smudge:
<< domain LGTM -- I don't mind the new naming. Was thinking about whether this should be a minor version bump instead of  a patch version bump, but I don't see how this would affect consumers of the code -- nobody should be referencing these classes directly, and the classes shouldn't appear in DJ handlers or other object-serializing contexts. So theoretically there are no API or behavior changes here for existing consumers.

RowanMcDonald:
Rebased to get github actions!

samandmoore (approved):
domainlgtm

samandmoore:
oh

<<platformlgtm

i am cool with this change. i tend to prefer the regular camelization of names like this anyway and Nathan said, this stuff should be unused by consumers.

nanda-prbot (approved):
Approved! :zap: :clap: :fried_egg:

PR #113, opened 2021-03-23
Introduce public url var that's shared with browser client

aburgel:
### Summary

For deployments where public traffic should run over a different network than service API calls, we're providing another environment variable that allows you to configure the publicly accessible TestTrack url. Now you can provide `TEST_TRACK_PUBLIC_API_URL` which will be passed along to the javascript client. By default, publicly accessible traffic will continue to use `TEST_TRACK_API_URL`.

/no-platform


RowanMcDonald:
<< domainLGTM! :shipit:

nanda-prbot (approved):
Approved! :avocado: :clap: :ghost:

PR #114, opened 2021-05-11
Introducing: ActiveJob support

smudge:
/platform @effron 

### Summary

This PR removes the **direct** dependency on `delayed_job_active_record` in favor of an indirect dependency via `activejob`, allowing downstream consumers to choose other queue adapters for the two background tasks that this library performs.

To use ActiveJob, this PR adds two new job classes:
- `AssignmentEventJob`, which is an ActiveJob-friendly rewrite of `NotifyAssignmentJob`
- `IdentifierCreationJob`, which replaces a usage of `.delay`

The DJ-compatible `NotifyAssignmentJob` class was left in place for compatibility with jobs that are still in the queue when this change rolls out, and will be removed in a future version. (It's just a PORO class with a "perform" method, so it's safe to keep around.)

### Other Information

- As a result of this change, this gem no longer had any reason to connect to a database, so I reduced its dependencies from `rails` (which depends on everything including `activerecord`) to just `railties`, `activemodel`, `sprockets-rails`, and `activejob` (the four rails-land dependencies that are actually required in order to use this gem's functionality). I also removed the `sqlite3` test dependency and any schema-related setup.

- The gem's test suite now uses the ActiveJob `:test` adapter, alleviating the need for a real adapter when running tests. (ActiveJob's internal serialization/deserialization/execution mechanisms don't vary in behavior across adapters, and we don't use any other adapter-specific features or test for anything relating to db transactions, so it didn't feel like we'd gain much by pulling in a non-test queue adapter during test runs.)

- In order to remove all direct references to `delayed_job`, I also cut out the code that attempts to load `delayed-plugins-airbrake` -- error reporting feels like something that consuming apps should configure for themselves. It looks like we might've been burned by a misconfiguration previously, so I left in some misconfiguration-detection code for anyone still relying on Airbrake, but I'd be happy to cut that out too. 🤷 

- **This gem does still ship a `delayed_job` plugin** (`TestTrack::DelayedJob::JobSessionPlugin`) but it was always fully optional -- never loaded by default -- and can still be optionally loaded! Existing usages will continue working as before.

### Given all of the above, this bumps the gem's version to 5.0.0

I could see a case for going with 4.1.0 -- theoretically if you're already using `delayed_job` and have properly configured it for use with `activejob`, then I guess you might not notice this change. But I didn't want to inadvertently break someone who _isn't yet using activejob_ since a misconfiguration there could cause the new jobs to run with an in-memory async adapter. I could go either way though.


app/jobs/test_track/assignment_event_job.rb
```diff
@@ -0,0 +1,71 @@
+module TestTrack
+  class AssignmentEventJob < TestTrack.job_base_class_name.constantize
+    attr_reader :visitor_id, :assignment
+
+    def perform(visitor_id:, assignment:)
+      raise "visitor_id must be present" if visitor_id.blank?
+      raise "assignment must be present" if assignment.blank?
+
+      @visitor_id = visitor_id
+      @assignment = build_assignment(visitor_id, assignment)
```
smudge, line 10: This `build_assignment` step is the one new thing about this class. The `TestTrack::Assignment` model is not serializable by ActiveJob, and rather than writing a custom serializer, I pass its fields through as a Hash and then reconstruct it inside of the job.

effron (commented):
looks great! a few minor questions

effron:
<<platformLGTM

lib/test_track.rb
```diff
@@ -20,6 +20,7 @@ module TestTrack
   BUILD_TIMESTAMP_REGEX = /\A\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d(.\d+)?([+-][0-2]\d:[0-5]\d|Z)\z/.freeze
 
   mattr_accessor :enabled_override, :app_name
+  mattr_accessor(:job_base_class_name) { 'ActiveJob::Base' }
```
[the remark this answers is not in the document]
samandmoore: yea i'm fine with this.

lib/test_track_rails_client/engine.rb
```diff
@@ -1,19 +1,5 @@
...
-  # no-op
-end
-
-unless defined?(Delayed::Plugins::Airbrake) && Delayed::Worker.plugins.include?(Delayed::Plugins::Airbrake)
-  begin
-    require 'delayed-plugins-airbrake'
-  rescue LoadError
-    # no-op
-  end
-end
```
samandmoore, line 15: i'm totally fine with removing this. we bundled it by default for a bit then when we stopped using airbrake internally we needed to make it optional and we ended up here.

vendor/gems/fakeable_her/fakeable_her.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency 'her'
-  spec.add_dependency 'rails', '>= 4.1', '< 5.2'
+  spec.add_dependency 'activesupport'
```
samandmoore, line 18: ha that's a funny change

samandmoore:
> Given all of the above, this bumps the gem's version to 5.0.0
> I could see a case for going with 4.1.0 -- theoretically if you're already using delayed_job and have properly configured it for use with activejob, then I guess you might not notice this change. But I didn't want to inadvertently break someone who isn't yet using activejob since a misconfiguration there could cause the new jobs to run with an in-memory async adapter. I could go either way though.

I'm fine with going to 5.0.0. Are there any small deprecations that we wanna clean up while we're here? I don't care that much but I feel like I remember some from the 4.0.0 release that we missed

smudge:
bump -- will also ask around re: things we can clean up in advance of 5.0.0

samandmoore:
domainlgtm

i'm cool with you shipping this without any other major version clean-up.

nanda-prbot (approved):
Approved! :nut_and_bolt: :dancers: :burrito:

PR #116, opened 2022-05-10
Fix CI workflows

smudge:
/no-platform

GH Actions no longer supports 16.04, so they spin forever.


HipsterBrown:
<< domainLGTM

nanda-prbot (approved):
Approved! :fries: :star2: :100:

PR #126, opened 2024-03-12
Set build timestamp initializer order

smudge:
### Summary

I noticed that some recent changes to my load order in a consuming app resulted in this initializer running too soon, before `Time.zone` was defined by Rails. This change forces it to always run _after_ `'active_support.initialize_time_zone'`.

/no-platform

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    test_track_rails_client (7.1.0)
+    test_track_rails_client (7.1.1)
       activejob (>= 6.0)
       activemodel (>= 6.0)
       faraday (>= 0.8)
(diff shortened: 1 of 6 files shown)


jmileham:
Is this initializer name we're hooking to stable across our target rails version range or could it be not effective in past or future versions?

smudge:
@jmileham it hasn't changed in [15 years](https://github.com/rails/rails/blame/dd3360e05e4909f2f0c74a624cccc2def688f828/activesupport/lib/active_support/railtie.rb#L18). Initializer names also get included in changelogs and _tend_ to be treated as part of the public API, though I wouldn't be surprised if the enforcement there is a little looser.

jmileham (approved):
Domain lgtm

PR #131, opened 2024-07-09
Prohibit feature flag

okasten:
### Summary
Adds a `prohibit_feature_flag` method that acts as the opposite of `require_feature_flag`. 
Example use case: the team is working on rolling out 2fa for a new app. In phase 1, you allow a user to disable their 2fa configuration. In phase 2, you want to require 2fa configuration behind a new feature flag. So you have a `disable_2fa_controller.rb` where you want to `require_feature_flag 2fa_enabled`, but `prohibit_feature_flag 2fa_required_enabled`, until the time comes where you can safely delete the entire controller at the end of your project.

### DIFF
--- a/app/controllers/concerns/test_track/controller.rb
+++ b/app/controllers/concerns/test_track/controller.rb
@@ -15,6 +15,12 @@ module TestTrack::Controller
         raise ActionController::RoutingError, 'Not Found' unless test_track_visitor.ab(feature_flag, context: self.class.name.underscore)
       end
     end
+
+    def prohibit_feature_flag(feature_flag, *args)
+      before_action(*args) do
+        raise ActionController::RoutingError, 'Not Found' if test_track_visitor.ab(feature_flag, context: self.class.name.underscore)
+      end
+    end
   end
 
   private
--- a/spec/controllers/concerns/test_track/test_trackable_controller_spec.rb
+++ b/spec/controllers/concerns/test_track/test_trackable_controller_spec.rb
@@ -6,6 +6,7 @@ RSpec.describe TestTrack::Controller do
   controller(ApplicationController) do
     include mixin
     require_feature_flag :cool_feature_enabled
+    prohibit_feature_flag :lame_feature_enabled
 
     self.test_track_identity = :current_clown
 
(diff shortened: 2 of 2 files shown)


jmileham:
Does this potentially orphan people mid flow? My mental model for require feature flag is that it hides new functionality but we generally will not want to hide a whole legacy controller until all users are done with flows. Switching an experience to 404 also is very hard to measure user impact because we generally will ignore 404s in error monitoring and alerting.

In context of use, is there a security or behavior problem this solves? I can see that maybe there's a "we actually need this" argument that we could provide guidance around, but generally "draining" use of legacy endpoints by having flags and client upgrades eliminate use gradually and then eventually removing the code for them might be fine without this? Is this to mitigate an assumption that it's safe to remove endpoints that might actually be still in use? Should we look for an alternative strategy for ensuring usage has drained before removing the route instead?

app/controllers/concerns/test_track/controller.rb
```diff
@@ -15,6 +15,12 @@ def require_feature_flag(feature_flag, *args)
         raise ActionController::RoutingError, 'Not Found' unless test_track_visitor.ab(feature_flag, context: self.class.name.underscore)
       end
     end
+
+    def prohibit_feature_flag(feature_flag, *args)
+      before_action(*args) do
+        raise ActionController::RoutingError, 'Not Found' if test_track_visitor.ab(feature_flag, context: self.class.name.underscore)
+      end
+    end
```
willlockwood, line 23: Have y'all considered that this could be the same method, but we could add an optional `variant:` kwarg to it, to relax the assumption we've made that they always want to match on the true variant? If this is functionality we want, maybe we want users to be explicit about which variant they want, and abandon that assumption. This would also allow for us to more easily expand to support non-traditional cases of `true_variant` not being `true` (which isn't currently supported here, from what I can tell).

```ruby
def require_feature_flag(feature_flag, *args, variant: true)
  before_action(*args) do
    raise ActionController::RoutingError, 'Not Found' unless test_track_visitor.ab(feature_flag, context: self.class.name.underscore) == variant
  end
end
```

Maybe even better to not give it a default, and introduce it as a breaking change that users are required to include a variant, but that may also just be a more straightforward interface for developers

smudge: Having a rock-solid and opinionated experience for true/false feature gating feels preferable to expanding the required interface for the rare cases when we want to gate on a variant other than "true." I'm not opposed to adding some extensibility here (especially in cases where we might want to gate an entire controller under a branch of an experiment) but my vote would be to default it to "true" for any true feature gates (i.e. if the name ends in `_enabled`).

okasten: I didn't consider that, but I do like the idea! I think I immediately went to a new method because it was so ingrained in my head that `require_feature_flag` meant it **had** to be `true`. I'm definitely open to expanding on `require_feature_flag` if that's what the majority deems preferable!

smudge:
We chatted offline a bit about how to ensure that the old variant of a feature is drained. The technique used roll a new feature out behind a gate doesn't quite work for turning a feature off, because the impact on UX is different; a user should see a 404 for a feature that they do not yet have access to, but a 404 is a poor experience for a feature they may be in the middle of using when the toggling happens.

This does not mean that there isn't generalizable API that could assist with this, but it would need to present a solution to the UX. For now, in cases where simply waiting for an old variant of a feature to "drain" is not sufficient, the best practice would be for end-developers to handle this experience manually in their code, via a validation error or warning/alert that can be customized to explain to the end user what just happened.

PR #132, opened 2024-07-23
fix: Prevent multiple identical instantiations of application identity within calls to `TestTrack.app_ab`

willlockwood:
### Summary

Earlier I noticed that we were hitting N+1s on calls to `TestTrack.app_ab` for the application visitor identity, the main fix for which is to memoize the instance of `Identity` we use for calls to `TestTrack.app_ab`. So, just adding in some logic that memoizes the identity across multiple calls (given that the app name doesn't change between calls).

### DIFF
--- a/app/models/test_track/application_identity.rb
+++ b/app/models/test_track/application_identity.rb
@@ -12,7 +12,7 @@ class TestTrack::ApplicationIdentity
   end
 
   def identity
-    @identity = @identity&.app_name == app_name ? @identity : Identity.new(app_name)
+    @identity ||= Identity.new(app_name)
   end
 
   class Identity
@@ -23,6 +23,8 @@ class TestTrack::ApplicationIdentity
     def initialize(app_name)
       @app_name = app_name
     end
+
+    private
 
     attr_reader :app_name
   end
--- a/app/models/test_track/application_identity.rb
+++ b/app/models/test_track/application_identity.rb
@@ -12,7 +12,7 @@ class TestTrack::ApplicationIdentity
   end
 
   def identity
-    Identity.new(app_name)
+    @identity = @identity&.app_name == app_name ? @identity : Identity.new(app_name)
   end
 
   class Identity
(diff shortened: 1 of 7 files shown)


app/models/test_track/application_identity.rb
```diff
@@ -12,7 +12,7 @@ def app_name
   end
 
   def identity
-    Identity.new(app_name)
+    @identity = @identity&.app_name == app_name ? @identity : Identity.new(app_name)
```
smudge, line 15:
```suggestion
    @identity ||= Identity.new(app_name)
```

If we're worried about `app_name` changing, we could make it so that it raises an exception if you try to reassign `TestTrack.app_name=` after a value has already been assigned... 🤔

app/models/test_track/application_identity.rb
```diff
@@ -24,8 +24,6 @@ def initialize(app_name)
       @app_name = app_name
     end
 
-    private
-
```
smudge, line 28: This should no longer be necessary if [this change](https://github.com/Betterment/test_track_rails_client/pull/132/files#r1688949530) is accepted.

app/models/test_track/application_identity.rb
willlockwood: So, this is what I originally had, but then memoizing the value broke a test that we had where the `TestTrack.app_name` gets set in a `before` block. Basically as your thing is written, after memoizing, resetting `TestTrack.app_name` has no effect, so the test that verifies what happens when the `app_name` is not set fails (it expects an error but gets none because `app_name` was set in a previously run test).

Any thoughts as to how I can write the test such that that's not an issue? Because if that's reasonably possible, then yes this is also what I'd prefer!

smudge: For the context of a test, we could:

```ruby
allow(TestTrack).to receive(:app_name).with(...)
```

willlockwood: OK cool got it so the problem was that the identity ivar was persisting across specs so I set it up so that it's cleared out before all specs. Made `app_name` private again, and now just using `@identity ||= ...` 👍🏻

smudge (approved):
domain LGTM && platform LGTM!

PR #144, opened 2025-01-30
Extract method for feature flag check

stevenjackson:
While working on a controller we wanted a fallback behavior where we could check another related association for a test track assignment if the visitor did not have direct access.

With this in place a controller can override the check when needed without completely replacing the before_action behavior

```
def feature_flagged?(feature_flag)
  super || current_user.organization.test_track_ab(feature_flag, context: self.class.name.underscore)
end
```


jmileham:
@smudge I think from a brief read of the ref'd retail PR discussion, this dips a toe into the zone that we want to address with parameterized splits a la split.io to get domain layer segmentation happening? Not sure whether this is a good thing to do til we're ready to bite that off or what, but your brain would be super helpful here.

